PR #18, opened 2023-06-07
Clean up opensource typeid code so we can publish it

loreto:
Clean up typeid code so we can publish it:
- Put spec under `typeid/typeid`. It will be published to `github.com/jetpack-io/typeid`
- Put go implementation under `typeid/typeid-go`. It will be published to `github.com/jetpack-io/typeid-go`
- Put sql implementation under `typeid/typeid-sql`. It will be published to `github.com/jetpack-io/typeid-sql`


typeid/typeid-go/README.md
```diff
@@ -0,0 +1,10 @@
+# TypeID Go
+### A golang implementation of [TypeIDs](https://github.com/jetpack-io/typeid)
+![License: Apache 2.0](https://img.shields.io/github/license/jetpack-io/typeid)
+
+TypeIDs are a modern, **type-safe**, globally unique identifier based on the upcoming
+UUIDv7 standard. They provide a ton of nice properties that make them a great choice
```
savil, line 6: curious, why uuidv7?

savil: okay, README below addresses it

savil (approved):

typeid/typeid-go/typeid.go
```diff
@@ -0,0 +1,62 @@
...
+	if tid.prefix == "" {
+		return tid.suffix
+	}
+	return tid.prefix + "_" + tid.Suffix()
+}
+
+func (tid TypeID) UUIDBytes() []byte {
+	b, err := base32.Decode(tid.suffix)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
loreto: Hehe, why yikes?

This error will never happen since the constructors will be the ones throwing an error, and guaranteeing once and object is constructed the data is valid and can be decoded. I see it as the way to make the compiler realize the invariant so that the return type is accurate.

(that said, if we do get a panic, I'll eat my words and change the signature, lol)

PR #19, opened 2023-06-07
Make it possible to publish opensource repos in subdirectories

loreto:
Change the github action we use to publish repos from the opensource monorepo, to support publishing repos in subdirectories.

With this change we'll be able to publish `typeid/typeid-go` to a `typeid-go` repo. It also makes it possible to specify a custom name that does *not* match the directory name for cases when that's needed (example `dir1/dir2:reponame`).

A follow up PR will use this github action to publish the typeid code.

### DIFF
--- a/action-publish-code/publish-code.sh
+++ b/action-publish-code/publish-code.sh
@@ -7,13 +7,13 @@ set -o errexit
 set -o nounset
 
 if [ "$#" = 0 ]; then
-	echo "usage: $0 <org>/<repo> <dir1> [<dirN>]..." 1>&2
+	echo "usage: $0 <org>/<repo> <dir1>[:<repo1>] [<dirN>[:<repoN>]]..." 1>&2
 	exit 1
 fi
 
 if [[ "$1" != *"/"* ]]; then
 	echo "error: first argument must be in the form <owner>/<repo>" 1>&2
-	echo "usage: $0 <org>/<repo> <dir1> [<dirN>]..." 1>&2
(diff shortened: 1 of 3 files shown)


action-publish-code/publish-code.sh
```diff
@@ -37,10 +37,16 @@ git clone -b main --single-branch "git@github.com:${org}/${origin_repo}.git" "${
...
 echo -e "\n"
 cd "${TMPDIR}/${origin_repo}"
-for dir in "$@"; do
-	echo "Validating $dir"
+for arg in "$@"; do
+	dir="${arg%:*}"
+	repo="${dir##*/}"
+	if [[ "$arg" != *":"* ]]; then
+		repo="${arg##*:}"
+	fi
```
savil, line 45: I trust you  :D

action-publish-code/publish-code.sh
```diff
@@ -55,13 +61,18 @@ done
 exit 0
 
 
-for repo in "$@"; do
+for arg in "$@"; do
+	dir="${arg%:*}"
+	repo="${dir##*/}"
+	if [[ "$arg" != *":"* ]]; then
+		repo="${arg##*:}"
```
savil, line 68: if this was golang, I'd suggest avoiding re-doing this work that was also done above.

But since its bash, copy-pasta may be easier :)

loreto: Hehe, yeah. FWIW, eventually I might rewrite in `go`. But it's working now so ....

PR #20, opened 2023-06-07
Clarify why the panic case won't trigger in the base32 package

loreto:
Clarify why the panic case won't trigger in the base32 package


savil (approved):
Thanks!

PR #24, opened 2023-06-13
Move typeid cli to typeid/ repo

loreto:
Move the typeid cli to the `typeid` repo, instead of `typeid-go`. This way, `typeid-go` will be solely for the go library. The top level `typeid` repo provides a cli tool to generates `typids` from the command-line. It happens to be written in go, but that's incidental.

This is a tiny PR that I wrote using Devbox Cloud to go through the whole experience.


savil (approved):
How does this work?

I would expect the import to need to be changed to `typeid-go`
https://github.com/jetpack-io/opensource/blob/0163961104cb4f035fba6d5c14f8b6fa20d4b1c8/typeid/typeid/main.go#L7

approving to unblock

loreto:
> How does this work?
> 
> I would expect the import to need to be changed to `typeid-go`
> 
> https://github.com/jetpack-io/opensource/blob/0163961104cb4f035fba6d5c14f8b6fa20d4b1c8/typeid/typeid/main.go#L7
> 
> approving to unblock

The name of the `repo` (and hence the directory) is `typeid-go`, but the name of the go module, as defined in its `go.mod` file is `go.jetpack.io/typeid`

Does that make sense?

PR #26, opened 2023-06-22
Implement a first version of TySON

loreto:
TySON is TypeScript for configuration files.

In the process of exploring how we could allow users to build packages using go, I ended up exploring whether we could use TypeScript as a backwards compatible extension of JSON (since that's the current format of our configuration files).

I ended up w/ TySON, and I really like it and think we should open-source it and get feedback from the community. A lot of the rationale for why I think it's a good idea is in the README (so please read it!), but a quick summary might be:
+ Backwards compatible with our existing configuration
+ Gives us comments, type safety, trailing commas, and functions when writing configuration
+ It's just TypeScript – we can use a lot of the existing ecosystem, like syntax highlighting and autocompletion on VSCode
   just works.
+ Built on the rock-solid `esbuild` that is widely adopted in the JS ecosystem. So I know the typescript support if first-class.

As last point I mention that I've been thinking about configs for a while know, including exploring HCL, Skylark, JSON5, writing our own language, etc, etc and I think this is the best approach I've run into so far.


tyson/internal/tsembed/eval.go
```diff
@@ -0,0 +1,16 @@
+package tsembed
+
+import "github.com/dop251/goja"
+
+func evalJS(code string) (goja.Value, error) {
+	vm := goja.New()
```
LucilleH, line 6: What's `vm`? `value map`? `virtual machine`?

LucilleH (approved):
I haven't been this excited about Typescript for a while 🤯

tyson/internal/tsembed/eval.go
loreto: "virtual machine" (as in the virtual machine that interprets the javascript)

PR #30, opened 2023-06-26
Update README.md

LucilleH:
Add additional examples. See markdown preview

### DIFF
--- a/tyson/README.md
+++ b/tyson/README.md
@@ -1,26 +1,68 @@
 # TySON 🥊
-### Use TypeScript as a configuration language
+### TypeScript as a Config Language
 
 # What is it?
 TySON (TypeScript Object Notation) is a subset of TypeScript, chosen to be useful as an embeddable configuration
-language that generates JSON. You can think of it as JSON + types + functions using
+language that generates JSON.
+You can think of TySON as **JSON + comments + types + basic logic** using
 TypeScript syntax. TySON files use the `.tson` extension.
 
-The goal is to make it possible for all major programming languages to read
-configuration written in TypeScript using native libraries.
+Here's a simple example.tson:
 
-The benefits of using TySON include:
-- **Type safety**: use TypeScript's type system to ensure that your configuration
-  is valid.
-- **Programmable**: you can define functions and modules that generate configuration
-  programmatically.
-- **Nicer Syntax**: unlike JSON, TypeScript supports comments, trailing commas,
-  and multi-line strings, in addition to types and functions. Unlike languages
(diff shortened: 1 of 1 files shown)


tyson/README.md
```diff
@@ -1,26 +1,68 @@
 # TySON 🥊
-### Use TypeScript as a configuration language
+### TypeScript as a Config Language
```
loreto, line 2: Spell out `Configuration`:
`TypeScript as a Configuration Language`

Because `Configuration Language` is the term being used in the space. See https://dhall-lang.org/ or https://jsonnet.org/

tyson/README.md
```diff
@@ -1,26 +1,68 @@
...
-- **Editor Support**: because TySON is a subset of TypeScript, your editor already
-  supports syntax highlighting, formatting and auto-completion for it. Simply
-  configure your editor to treat `.tson` files as TypeScript files.
+Here's a simple example.tson:
+
+```
+// example.tson
+export default {
+  // Comments
+  string_field: 'string',
```
loreto, line 16: Could you have one of the fields be a multiline string using template literals?

tyson/README.md
```diff
@@ -1,26 +1,68 @@
...
+  number_field: 123,
+  boolean_field: true,
+  array_field: [1, 2, 3], // Add more comments.
+  object_field: {
+    nested_field: "nested",
+  }
+}
+```
+
```
loreto, line 25: I would add a line here: "It evaluates to the following JSON" and the write out the equivalent JSON

tyson/README.md
```diff
@@ -1,26 +1,68 @@
...
+### TypeScript as a Config Language
 
 # What is it?
 TySON (TypeScript Object Notation) is a subset of TypeScript, chosen to be useful as an embeddable configuration
-language that generates JSON. You can think of it as JSON + types + functions using
+language that generates JSON.
+You can think of TySON as **JSON + comments + types + basic logic** using
 TypeScript syntax. TySON files use the `.tson` extension.
 
-The goal is to make it possible for all major programming languages to read
```
loreto, line 9: I don't want to fully get rid of this sentence. Since TySON is in some ways just TypeScript, I think an objection to it can be "why do you even need TySON, just use TypeScript directly" – the idea of making it accessible from several host languages without a direct dependency in the JS ecosystem is valuable.

Is there a different place where we could keep this sentence? Or something similar?

LucilleH: Oh, did I delete this sentence? That was by mistake. Adding it back...

loreto (approved):
LGTM

Although, one more request, could we have syntax highlighting in the code examples you added?

I think it's a matter of using:
```
```typescript

```json
```

PR #32, opened 2023-06-26
Test and build go modules in opensource monorepo

loreto:
Add a github action that tests, and builds every go module in the opensource monorepo as part of CICD (including `tyson` and `typeid`).


LucilleH (approved):
General QQ: in the event that people start contributing to TySON opensource repo directly, what happens? Can we then detach the tyson repo from this one?

loreto:
> General QQ: in the event that people start contributing to TySON opensource repo directly, what happens? Can we then detach the tyson repo from this one?

We can, yes – the setup is a two way door and the the repo can easily be "spun off". That said, I am interested in experimenting with the monorepo format, and adding `CONTRIBUTING` instructions to the individual projects that explain how to contribute back via the monorepo.

PR #33, opened 2023-06-26
Update README.md

LucilleH:
Add devbox and jetpack.io links

### DIFF
--- a/tyson/README.md
+++ b/tyson/README.md
@@ -48,6 +48,9 @@ The above evaluates to the following JSON:
 }
 ```
 
+TySON was originally developed by [jetpack.io](https://www.jetpack.io). We are exploring 
+using it as a configuration language for [Devbox](https://github.com/jetpack-io/devbox).
+
 ## Benefits of using TySON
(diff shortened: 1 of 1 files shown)


tyson/README.md
```diff
@@ -87,8 +90,12 @@ Simply configure your editor to treat `.tson` files as TypeScript files.
 
 
 # Why?
-Almost all developer tools require some form of configuration. In our opinion,
-an ideal configuration language should be:
+As we are developing [Devbox](https://github.com/jetpack-io/devbox), many users ask
```
loreto, line 93: I like the first paragraph ("TySON was originally developed ..."), but this one for me feels like it detracts from marketing TySON as a general purpose configuration language. I worry people will start getting the impression that it's a Devbox specific language, and they can't / shouldn't adopt it for their own projects.

Q: Are you ok w/ just the first paragraph and revering this one? If yes, let's do that. If no, let me know, and I can put some thought on how we incorporate what you have here in a way that doesn't change the overall message.

LucilleH: I can remove

loreto (approved):
LGTM

PR #34, opened 2023-06-26
Add devbox, lint markdown

Lagoja:


loreto:
Looks good, although I'm assuming you're not intending to revert the changes the README that Lucille made and you just need to merge the latest `main`

tyson/README.md
```diff
@@ -90,29 +94,32 @@ Simply configure your editor to treat `.tson` files as TypeScript files.
...
 or TOML, but they each have drawbacks:
-+ **JSON**: doesn't support comments, trailing commas, or multi-line strings.
-+ **YAML**: has an ambigous syntax. For example the token `no` is interpreted
+
+- **JSON**: doesn't support comments, trailing commas, or multi-line strings.
+- **YAML**: has an ambigous syntax. For example the token `no` is interpreted
   as a boolean, often in cases where you want it to be a string. See
   https://noyaml.com/ for more examples.
-+ **TOML**: Gets unwidely when there's multiple levels of nesting.
+- **TOML**: Gets unwidely when there's multiple levels of nesting.
```
LucilleH, line 122: What's the difference between `+` and `-`?

Lagoja:
> Looks good, although I'm assuming you're not intending to revert the changes the README that Lucille made and you just need to merge the latest `main`

I accepted all her changes, just merged in the lint modifications

tyson/README.md
Lagoja: I think `*` or `-` are the more standard Markdown bullets

tyson/README.md
```diff
@@ -1,7 +1,11 @@
 # TySON 🥊
-### TypeScript as a Configuration Language
+
+**Use TypeScript as a configuration language.**
```
loreto, line 3: This is what I was seeing as revert:
1. We switched the language to `TypeScript as a Configuration Language`
2. I actually do prefer to use `###` here rather than `**` – it's what we had been doing in the Devbox repo as well. I both like the line that the header creates, and I think from an SEO perspective it's preferable for the HTML to end w/ a header tag.

PR #42, opened 2023-06-28
Fix the way we publish tags to subrepos

loreto:
Fix the way our github action publishes tags to subrepos. Needed so that we can successfully create releases in the subrepos.

### DIFF
--- a/action-publish-code/publish-code.sh
+++ b/action-publish-code/publish-code.sh
@@ -69,9 +69,14 @@ for arg in "$@"; do
 	set -o xtrace # Print commands as they are executed
 
 	# Remove everything we don't want from the source repo:
+ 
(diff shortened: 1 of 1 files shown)


savil (approved):
not really following, but approving to unblock

action-publish-code/publish-code.sh
```diff
@@ -69,9 +69,14 @@ for arg in "$@"; do
 	set -o xtrace # Print commands as they are executed
 
 	# Remove everything we don't want from the source repo:
+ 
+	# Only keep the tags belonging to the repo we care about
+	git tag -d $(git tag -l | grep -v "${repo}/*")
```
savil, line 74: oh this one makes sense 👍🏾

PR #47, opened 2023-06-29
Adds string validation before creating a typeid via FromString

fernandezpablo85:
Migrated from https://github.com/jetpack-io/typeid-go/pull/1

Compared to the original PR this one tolerates typeids without prefix. It also considers a `[prefix]_[suffix]_[random]` invalid, which I'm assuming it's the right thing to do, specially to support symmetry when going back and forth from strings to typeids.


loreto (approved):
LGTM

PR #48, opened 2023-06-29
Add unit tests to typeid go implementation

loreto:
More unit testing was one of the main request from the HackerNews discussion


typeid/typeid-go/typeid_test.go
```diff
@@ -0,0 +1,136 @@
...
+	fmt.Printf("New typeid without prefix: %s\n", tid)
+}
+
+func ExampleFromString() {
+	tid := Must(FromString("prefix_00041061050r3gg28a1c60t3gf"))
+	fmt.Printf("Prefix: %s\nSuffix: %s\n", tid.Type(), tid.suffix)
+	// Output:
+	// Prefix: prefix
+	// Suffix: 00041061050r3gg28a1c60t3gf
+}
```
savil, line 24: Why these? Could we wrap them in a test that checks the returned values are as expected?

typeid/typeid-go/typeid.go
```diff
@@ -112,10 +112,18 @@ func Must(tid TypeID, err error) TypeID {
 }
 
 func validatePrefix(prefix string) error {
+	if prefix == "" {
+		return nil
+	}
+
+	if len(prefix) > 63 {
```
savil, line 119: 👍🏾

savil (approved):

typeid/typeid-go/typeid_test.go
loreto: No, these are examples, which go treats differently so that they can be rendered as part of the documentation. If I turn them into tests, they will no longer appear as examples in the doc.

PR #49, opened 2023-06-29
Add TypeScript version of typeid to opensource repo

loreto:
For now I've copied the code we are using in our frontend. Later, once we publish this package in npm, we should remove the frontend version and just rely on this one.


savil (approved):
Sounds good. Would be great to add a `devbox.json` to this directory as well.

PR #62, opened 2023-06-30
add github.com/broothie/typeid-ruby to `README.md`

broothie:
Adding a Ruby implementation of typeid, [typeid-ruby](https://github.com/broothie/typeid-ruby). It's been tested against the provided `valid.yml` and `invalid.yml` files. It does not come with a CLI, not sure if that's a requirement!


loreto:
No CLI required. In fact, I think most people should just use the official CLI, and then use the libraries for when they want to use it in an application. Taking a look.

loreto (approved):
LGTM

PR #63, opened 2023-07-01
update rust type-safe-id compliance

conradludgate:
Test suite here: https://github.com/conradludgate/type-safe-id/blob/main/tests/spec.rs.
Fuzzing has also been configured and run for a large amount of time (10 minutes).


loreto (approved):
LGTM

PR #66, opened 2023-07-02
Add TypeID Java implementation to README

fxlae:
There is a [spec test class](https://github.com/fxlae/typeid-java/blob/main/src/test/java/de/fxlae/typeid/SpecTest.java) that specifically uses the two files (valid.yml/invalid.yml) for validation against the specification. I marked the entry in the table as "validated" and hope that's ok, if not please just let me know.


loreto (approved):
LGTM

PR #68, opened 2023-07-03
fix(typeid): correct that type is a prefix in spec overview

JanWennrich:
The spec overview incorrectly stated that the type is a suffix.  
The type is a prefix, as stated in the spec chapter "Type prefix".


loreto:
Thanks for the fix!

PR #69, opened 2023-07-03
Add `BombenProdukt/typeid` for `PHP` to `README.md`

faustbrian:
I've begun an implementation for PHP, which you can find at https://github.com/BombenProdukt/typeid. The encoding and decoding implementation is based on https://github.com/jetpack-io/typeid-js. At present, it passes all encodings as per https://github.com/jetpack-io/typeid/blob/main/spec/valid.yml. We are running the tests directly against the `valid.yml` and `invalid.yml` files to promptly detect any failures. ~Tests for the `invalid.yml` file are still missing, but I plan to add them in the coming days as time allows.~


loreto (approved):
LGTM

PR #70, opened 2023-07-03
Statically typed version of TypeIDs using generics

loreto:
Cleans up the godocs for the `typeid` go module.
Adds a statically typed version using generics.

### DIFF
--- a/typeid/typeid-go/README.md
+++ b/typeid/typeid-go/README.md
@@ -43,7 +43,7 @@ import (
 )
 
 func example() {
-  tid := typeid.New[UserIDType]()
+  tid := typeid.New[UserID]()
   fmt.Println(tid)
 }
 ```
--- a/typeid/typeid-go/README.md
+++ b/typeid/typeid-go/README.md
@@ -30,8 +30,9 @@ import (
   typeid "go.jetpack.io/typeid/typed"
 )
 
-type UserIDType struct{}
-func (UserIDType) Type() string { return "user" }
+type userPrefix struct{}
+func (userPrefix) Type() string { return "user" }
+type UserID struct { typeid.TypeID[userPrefix] }
 ```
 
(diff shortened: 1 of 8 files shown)


typeid/typeid-go/README.md
```diff
@@ -1,10 +1,63 @@
...
+
+To use it, first define your TypeID types:
+
+```go
+import (
+  typeid "go.jetpack.io/typeid/typed"
+)
+
+type UserIDType struct{}
+func (UserIDType) Type() string { return "user" }
```
mikeland73, line 34: This is cool, but should we recommend it? It makes stuff more verbose which makes the library a bit harder to adopt.

typeid/typeid-go/typed/typed.go
```diff
@@ -0,0 +1,115 @@
...
+		// in case, construct a "nil" id of the given type.
+		return Nil[T](), err
+	}
+	return (TypeID[T])(tid), nil
+}
+
+// Nil returns the null typeid of the given type.
+func Nil[T IDType]() TypeID[T] {
+	var prefix T
+	return TypeID[T](untyped.Must(untyped.From(prefix.Type(), "00000000000000000000000000")))
```
mikeland73, line 35: Can this be 

```golang
return TypeID[T]{}
``` 

or maybe:
```golang
return TypeID[T]{suffix: untyped.NIL.Suffix()}
``` 

?

It would be missing the prefix but is that something that makes sense in an error call?  (If we are telling users to ignore the value, having it contain prefix data could be worse). In a similar vein, I assume initializing it with `00000000000000000000000000` (both here and untyped) is just so that functions that require a suffix value won't break, but is that a property we want when functions return non-nil errors.

Don't feel strongly and not a blocker, just sharing thoughts.

mikeland73 (approved):

PR #72, opened 2023-07-04
[tyson] Implicitly export a single top-level object literal

loreto:
Adds a transformation, via a esbuild plugin, that implicitly exports top-level object literals in a .tson file when there are no other expressions in the file. This makes it possible to read `.json` files (which don't have a `default export` declaration) as valid TySON.


tyson/internal/tsembed/plugin.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return api.OnLoadResult{}, err
+	}
+
+	offset := findImplicitExport(original)
+	var builder strings.Builder
+
+	if offset != -1 {
+		builder.Write(original[:offset])
+		builder.WriteString("export default ")
```
LucilleH, line 33: Nice 😄

LucilleH (approved):

PR #74, opened 2023-07-05
Add Zig implementation to README

tensorush:
I've implemented TypeID in Zig and successfully tested it against the provided test data.


loreto:
@tensorush Thanks for contributing!

Take a look at https://github.com/jetpack-io/opensource/pull/75 since we're planning to add a few two more test cases (one in the valid file and one in the invalid file) to catch 128-bit overflows. You can go ahead and add those to your implementation.

loreto (approved):
LGTM

tensorush:
@loreto Cool, I've added those tests as well.

PR #75, opened 2023-07-05
[typeid] Enforce that the first suffix character is in the 0-7 range

loreto:
Change spec and implementations to enforce that the first suffix character should be in the `0-7` range. This is because
26 characters in base32 encode 130-bits, but UUIDs are only 128-bits. In order to guarantee that there are no
overflow errors and that the `typeid <-> uuid` mapping is a bijective function,  the maximum possible suffix is `7zzzzzzzzzzzzzzzzzzzzzzzzz`.

This was first reported by @fxlae in https://github.com/jetpack-io/typeid/issues/20

This PR:
1. Updates the spec
2. Updates the test data files provided by the spec
3. Updates our `go` and `typescript` implementations

### DIFF
--- a/typeid/typeid/spec/invalid.yml
+++ b/typeid/typeid/spec/invalid.yml
@@ -4,7 +4,7 @@
 # Each example contains an invalid TypeID string. Implementations are expected
 # to throw an error when attempting to parse/validate these strings.
 #
-# Last updated: 2023-06-29
+# Last updated: 2023-07-05
 
 - name: prefix-uppercase
   typeid: "PREFIX_00000000000000000000000000"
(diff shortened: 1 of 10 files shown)


akhundMurad (approved):

typeid/typeid/spec/invalid.yml
```diff
@@ -81,3 +81,8 @@
   # This example would be valid if we were using the crockford hyphenation rules
   typeid: "prefix_123456789-0123456789-0123456"
   description: "The suffix can't ignore hyphens as in the crockford encoding"
+
+- name: suffix-overflow
+  # This is the first suffix that overflows into 129 bits
+  typeid: "prefix_8zzzzzzzzzzzzzzzzzzzzzzzzz"
+  description: "The should encode at most 128-bits"
```
broothie, line 88:
```suggestion
  description: "The suffix should encode at most 128-bits"
```

PR #76, opened 2023-07-06
add java impl

softprops:
follow up on https://github.com/jetpack-io/typeid/pull/21#issuecomment-1622567164

moved pr here as well as updated to pass against the [new specs](https://github.com/softprops/typeid-java/commit/266bd10d9dbd447a7c8d1edc5e97e221dcf90618)


loreto (approved):
LGTM

PR #82, opened 2023-07-06
Update invalid.yml description

broothie:
- Added a script to update all spec-based files after a change to `typeid/typeid/spec/invalid.yml` is made.
- Updated spec files

### DIFF
--- a/typeid/typeid/scripts/update_specs.sh
+++ b/typeid/typeid/scripts/update_specs.sh
@@ -1,0 +1,13 @@
+#! env bash
+
+echo "Convert to JSON"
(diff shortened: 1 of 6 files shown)


typeid/typeid/scripts/update_specs.sh
```diff
@@ -0,0 +1,13 @@
+#! env bash
+
+echo "Convert to JSON"
+yq eval typeid/typeid/spec/invalid.yml --tojson > typeid/typeid/spec/invalid.json
```
loreto, line 4: Could you add similar commands for `valid.yml`?

loreto (approved):
LGTM

This looks great, just one comment: could you update the script you added to also update `valid.yml`?

PR #86, opened 2023-07-07
Add validation to Haskell implementation of typeid

MMZK1526:
The test spec is adapted from the [original test in Go](https://github.com/jetpack-io/typeid-go/blob/main/typeid_test.go).

Here's the successful CI: https://github.com/MMZK1526/mmzk-typeid/actions/runs/5490979657/jobs/10007064981


loreto (approved):
lgtm

PR #88, opened 2023-07-09
Implement TypeID prefix narrowing

xprnio:


LucilleH (commented):
I'm thinking if it make sense to change the default `TypeID`, as oppose to creating another `NarrowTypeID` type? As in, upon typeid construction, returns `TypeID<T>`?

Or you have thought about it, and the current way works better?

loreto:
> First off, this is awesome! Thank you!
> 
> I'm thinking if it make sense to change the default `TypeID`, as oppose to creating another `NarrowTypeID` type? As in, upon typeid construction, returns `TypeID<T>`?
> 
> Or you have thought about it, and the current way works better?

I also like that idea of `TypeID` itself returning `TypeID<T>`. @xprnio Would that work for your use case?

loreto:
Closing since this is now implemented via https://github.com/jetpack-io/opensource/pull/91
@xprnio Thanks so much for the suggestion!

xprnio:
Sorry, I forgot to check in with this :D No problem, I really needed it myself and that was the solution I came up with that would be backwards-compatible or wouldn't force the type-safety (since I did not know whether that was wanted or not). But glad to see it was pulled in, thank you  :D

PR #89, opened 2023-07-10
[typeid-sql] Add encoding/decoding functions

loreto:
- Add functions to encode and decode typeids from SQL
- Add supabase as a way to develop and test changes
- Add unit tests using pgTAP


LucilleH (approved):
Whoa!

PR #93, opened 2023-07-13
Add the strong_id crate as a Rust implementation of TypeID

johnnynotsolucky:
I have a Rust crate which implements TypeID-like behavior for unsigned ints and UUIDs. It is compliant with the TypeID spec when the "typeid" feature is enabled, which is the same as enabling both "uuid" and "uuid-v7" features.

I've included a dedicated crate which pulls the spec files and tests encoding/decoding the values: https://github.com/johnnynotsolucky/strong_id/tree/main/tests/typeid_spec

There is also a github workflow which runs on a 6 hour cron to confirm that the implementation is still valid: https://github.com/johnnynotsolucky/strong_id/blob/main/.github/workflows/typeid_spec.yaml

This crate takes a different approach in that it uses macros to generate TypeIDs so that we can have compile-time validation of prefix rules.


loreto (approved):
@johnnynotsolucky Thanks so much for contributing this implementation.

LGTM!

PR #98, opened 2023-07-15
Update date of Swift libs specs validation

Frizlab:
In response to this https://github.com/jetpack-io/typeid/pull/25#issuecomment-1631437351.


loreto (approved):
LGTM

PR #99, opened 2023-07-17
Adds in native typeid equality overloading

JasonYao:
Hey, I came across some behavior in a Java SpringBoot + postgres setup, and found what looks like a pretty generic solution that could be handy for other people.

Essentially, I needed the ability to query in java for the DB records by a given typeid, and instead of needing to drop down into the raw `databaseclient` to call the `typeid_parse` stored procedure that's already defined here, I could define an operator overload instead, which would allow me to just query a given DB table by the serialized type_id instead.

I added a comment in the actual `.sql` file, but let me know if you'd like me to make any changes to this PR.

### DIFF
--- a/typeid/typeid-sql/sql/03_typeid.sql
+++ b/typeid/typeid-sql/sql/03_typeid.sql
@@ -95,3 +95,39 @@ end
 $$
 language plpgsql
 immutable;
+
+-- Enables direct textual equality checks. This enables direct querying in pSQL without
+-- having to have clients know about db column internals- e.g. using the users table
+-- example in example.sql:
+--
+-- Query:
+-- SELECT * FROM users u WHERE u.id = 'user_01h455vb4pex5vsknk084sn02q'
+--
+-- Result:
+-- "(user,018962e7-3a6d-7290-b088-5c4e3bdf918c)",Ben Bitdiddle,ben@bitdiddle.com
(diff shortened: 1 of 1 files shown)


JasonYao:
No worries! Added in a quick `05_operator_test.sql` test along with an optional instruction step in the README, though I'm having some trouble running the tests since it looks like some setup functions might not have been committed yet- could you give the test file above a try and let me know if there's any issues, or share what the test configs should be so I can try running it on my local?

typeid/typeid-sql/supabase/tests/05_operator_test.sql
```diff
@@ -0,0 +1,27 @@
+-- Start transaction and plan the tests.
+BEGIN;
+SELECT plan(30);
```
JasonYao, line 3: I'm guessing that this int value is a seed so it enables the tests to run independently of one another even though they're using the same table name, but if not and there's some additional semantic meaning to it, let me know and I can change this to whichever plan value that you'd like

loreto: The number in `plan` is the number of tests that this test file is going to run; in this case `2`. I've gone ahead and updated it to `plan(2)`

loreto (approved):
LGTM

PR #102, opened 2023-07-19
[WIP] Install binary releases from Github

loreto:
Beginnings of a library that installs binary releases directly from Github. I'm not done with it yet, but I'm worried that if I don't send a PR now, it'll start getting too big. Let me know if we should try to check this in, and I can then continue in another PR, or if we use this PR as a "checkpoint" to do a quick review, but then I continue adding to it and get a second review when it's done.

A couple of notes:
+ I've created a `sandbox` directory in our `opensource` repo and added it there. The point of the `sandbox` is to contain code we're still experimenting with, and we might still choose to scrap or completely restructure later. For example, we might choose to embed this entire library inside of `devbox`, and delete it from `sandbox`
+ My current intention is to write this library so that we can do two things with it:
  1. Opensource a`launcher` and/or `installer` that replaces the bash scripts we've used in the past (we might still need one bash script to bootstrap, but the bulk of the code would now be go)
  1. Use the library within devbox to enable installation of binary packages there.
+ I'm not settled on the clis I'm including with the library. Consider those testing clis more than anything else.


mikeland73 (approved):
Merge it. No need to make early development more complicated than it should be.

PR #103, opened 2023-07-21
[runx] Add support for automatically running commands after pkg installation

loreto:
This PR does mainly two things:
1. Improves the heuristics for how we install/extract releases from github.
2. Introduces the notation that makes it possible to "execute" a binary from a package that gets installed on demand:
`runx +jetpack-io/tyson tyson`


sandbox/runx/impl/registry/registry.go
```diff
@@ -57,11 +125,20 @@ func (r *Registry) resolveVersion(ref types.PkgRef) (types.PkgRef, error) {
...
-	if len(releases) == 0 {
+	found := false
+	latestVersion := "latest"
+	for _, release := range releases {
+		if !release.Draft && !release.Prerelease {
+			found = true
+			latestVersion = release.TagName
+			break
+		}
+	}
```
mikeland73, line 136: If nothing is found, return pre-release?

mikeland73 (approved):

sandbox/runx/impl/registry/registry.go
loreto: Done

PR #105, opened 2023-08-21
[envsec] Add auth commands

mikeland73:
## Summary

This PR adds auth commands (mostly for testing, we can decide what the keep later). 

It adds the following common libraries from https://github.com/jetpack-io/devbox to a new `/pkg` directory:

* auth
* cuecfg
* envir (partially)
* usererr
* xdg

Most of these were moved as-is or with minimal changes. I moved them in order to get `auth` working as quickly as possible, but we might not need them long term. If we do want to keep some/all of these we can make devbox point to these instead.

## How was it tested?

```bash
envsec auth login
envsec auth whoami
```


mohsenari (approved):

loreto (approved):
LGTM

PR #107, opened 2023-08-22
Simplify dependencies for auth

loreto:
## Summary
Remove dependencies to `xdg`, `usererr`, and `cuecfg` from the authentication code so that we can refactor it into its own go module that we can opensource separately. 

A future PR will turn it into its own go module

## How was it tested?
`go test`


mikeland73 (approved):

mikeland73:
Is the plan to turn `auth` into own module?

PR #108, opened 2023-08-22
[envsec] Remove global flags

mikeland73:
## Summary

This turns previously global flags into per-command flags. It is needed for 2 reasons:

* Some commands don't need these flags
* Once we hook up jetpack cloud these values will be set automatically so we abstract away the envsec configuration.

Unrelated improvements:

* Moved CICD to use devbox
* Added lint to the entire repo (thanks ChatGPT for bash magic)

## How was it tested?

Built envsec:

```bash
envsec auth login
envsec auth whoami
envsec
envsec download --help
```

on top level of repo:

```bash
devbox run go-mod-sync
devbox run build
devbox run lint
devbox run test
```


mikeland73:
Bad rebase. Converting to draft until fixed

devbox.json
```diff
@@ -0,0 +1,20 @@
...
+    "golangci-lint@latest"
+  ],
+  "shell": {
+    "init_hook": [
+      "echo 'Welcome to devbox!' > /dev/null"
+    ],
+    "scripts": {
+      "build": "find . -type f -name go.mod -maxdepth 4 -execdir go build -v ./... \\;",
+      "go-mod-sync": [
+        "find . -maxdepth 4 -type f -name go.mod -execdir go mod tidy \\;",
```
loreto, line 13: Nit: keep the flags in the same order as in the other scripts?

envsec/.golangci.yml
```diff
@@ -1,4 +1,5 @@
 # TODO(Landau): Turn on all linters
+# TODO(Landau): move this to top level of opensource?
```
loreto, line 2: Yes, we should do that

loreto (approved):
LGTM

PR #109, opened 2023-08-24
[envsec] Implement init command

mikeland73:
## Summary

Init calls an API (stubbed) to fetch a new project ID. If ID already exists it prints error and displays the ID.

unrelated: moved all internal packages into `/internal`

## How was it tested?

envsec init


envsec/internal/jetcloud/client.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"os"
+
+	"go.jetpack.io/envsec/internal/auth"
+	typeid "go.jetpack.io/typeid/typed"
+)
+
+type projectPrefix struct{}
+
+func (projectPrefix) Type() string { return "project" }
```
loreto, line 13: In the DB, are we going to have a single projects table for all of our products or is the projects table specific to envsec? If specific to envsec, maybe we want `envsec_project`? (we'll have to change typeid to allow for `_` in the prefix, but I've been wanting to do that for a while)

envsec/internal/jetcloud/project.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"go.jetpack.io/envsec/internal/auth"
+)
+
+const dirName = ".jetpack"
```
loreto, line 12: Given that we're going to change the company name, will this mean we'll have to migrate the data in the directory?

loreto (approved):
LGTM

envsec/internal/jetcloud/client.go
LucilleH: Can we do `proj_` instead? Also, I would prefer that the project id format is consistent across all of our products

envsec/internal/jetcloud/git.go
```diff
@@ -0,0 +1,31 @@
...
+func createGitIgnore(wd string) error {
+	gitIgnorePath := filepath.Join(wd, dirName, ".gitignore")
+	return os.WriteFile(gitIgnorePath, []byte("*"), 0600)
+}
+
+func gitRepoURL(wd string) (string, error) {
+	cmd := exec.Command("git", "config", "--get", "remote.origin.url")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", err
```
LucilleH, line 19: What happens if the user does not have `git` installed?

envsec/internal/jetcloud/git.go
```diff
@@ -0,0 +1,31 @@
...
+	"strings"
+)
+
+func createGitIgnore(wd string) error {
+	gitIgnorePath := filepath.Join(wd, dirName, ".gitignore")
+	return os.WriteFile(gitIgnorePath, []byte("*"), 0600)
+}
+
+func gitRepoURL(wd string) (string, error) {
+	cmd := exec.Command("git", "config", "--get", "remote.origin.url")
```
LucilleH, line 16: Is this going to be different for `ssh` vs `https` method, based on how the user cloned the repo?

If so, are we doing the parsing/unification in the api?

envsec/internal/jetcloud/git.go
```diff
@@ -0,0 +1,31 @@
...
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
+func gitSubdirectory(wd string) (string, error) {
+	cmd := exec.Command("git", "rev-parse", "--show-prefix")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", err
```
LucilleH, line 28: Same thing, what happens if the user does not have `git` installed?

envsec/internal/jetcloud/git.go
mikeland73: if no git, we ignore errors and we pass empty strings for both repo and subdirectory. In this case the API should return a new ID.

envsec/internal/jetcloud/git.go
mikeland73: error is ignored, we se empty string

mikeland73:
> Can we do proj_ instead? Also, I would prefer that the project id format is consistent across all of our products

Yep of course.

mikeland73:
@LucilleH @loreto made requested changes, but there are still some open questions. 

* name of config directory. (I can just use `.envsec`, but it's a pain to use different name for each tool.)
* Name of typeid prefix (currently proj, we can change this.)

loreto:
> @LucilleH @loreto made requested changes, but there are still some open questions.
> 
> * name of config directory. (I can just use `.envsec`, but it's a pain to use different name for each tool.)
> * Name of typeid prefix (currently proj, we can change this.)

For dogfooding let's just move forward as is. But we should figure out the final dir name before opening up to users. Maybe that's coupled with what the new name of the company will be, I think I'll hear from the legal team pretty soon.

PR #110, opened 2023-08-24
Implement auth library using OIDC

loreto:
## Summary
First pass at implementing a standalone authentication library for command line tools. The idea is to open source it, in addition to using it in our own tools. The library implements the authentication flow with PCKE using OIDC.

At the moment this just has the basic flow, but I still need to add:
- Storage or caching of tokens
- Spinning up a server locally to listen for the flow's callback


savil:
is it "PCKE for OIDC", or PKCE: https://oauth.net/2/pkce/ ?

sandbox/auth/internal/pcke/pcke.go
```diff
@@ -0,0 +1,68 @@
...
+//
+// The standard go library for oauth2 does not support PCKE, but a proposal
+// has been accepted to add it in a future version of go. We copy the code from
+// the proposal so that it's easy to switch to the standard library when it's
+// available.
+//
+// Issue: https://github.com/golang/oauth2/issues/603
+// Proposal: https://github.com/golang/go/issues/59835
+
+package pcke
```
savil, line 12: I think needs to be pkce :)
https://go-review.googlesource.com/c/oauth2/+/463979

PR #111, opened 2023-08-25
[typeid] Add a T-SQL implementation in the README.md

titouancreach:
Hello,

We added an T-SQL naive implementation (allowing to migrate, create default constraint etc) in SQL Server.
It respects the specs (we can decode it using your CLI)

Regards


loreto:
Thanks so much for contributing!

loreto (approved):
LGTM

PR #112, opened 2023-08-26
[envsec] Added aws token exchange and fetching ssm params

mohsenari:
## Summary
Added first draft code for authenticating with aws cognito and fetching parameters from an `org_id`

The production ready code needs a bit more change since this one has hardcoded identity pool ids that I created via aws console UI. But we can review and merge this for testing before solidifying terraform changes.

## How was it tested?
- `devbox run build`
- `devbox auth login`
- `envsec init`
- `envsec set testkey=testvalue`
- `envsec ls`


envsec/internal/auth/aws.go
```diff
@@ -0,0 +1,155 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/MicahParks/keyfunc/v2"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/cognitoidentity"
+	"github.com/aws/aws-sdk-go/service/ssm"
```
mikeland73, line 14: These are all v1, make sure you only use v2.

envsec/internal/envcli/test.go
mikeland73, line 1: I think this is OK to keep in your own branch but not sure we should put it into main.

envsec/debug/debug.go
mikeland73, line 1: put into `/envsec/internal/debug`

envsec/internal/auth/aws.go
mikeland73, line 1: I think `auth` may be turned into own package that only handles user auth and not federation. Maybe move to `envsec/internal/awsfed` ? 

(we might eventually want to create a federation interface)

envsec/internal/auth/aws.go
```diff
@@ -0,0 +1,155 @@
...
+	"go.jetpack.io/envsec/debug"
+)
+
+// for valiation of JWT (can be hardcoded to avoid fetching)
+const jwksURL = `https://jetpack-io.us.auth0.com/.well-known/jwks.json`
+const awsRegion = "us-east-1"
+
+// have to be var instead of const so that their &address can be passed
+var accountId = "984256416385"
+var identityPoolId = "us-east-1:da3c3c71-61c7-4f7c-8e3d-3770e9b61379"
```
mikeland73, line 25: One thing I find helpful is to declare a config struct and then put hard coded stuff there. That way once we stop hardcoding it's easy to replace with env vars or build vars etc.

envsec/internal/auth/aws.go
```diff
@@ -0,0 +1,155 @@
...
+
+	byteContent, err := io.ReadAll(authFile)
+	if err != nil {
+		return "", err
+	}
+
+	var result auth
+	json.Unmarshal([]byte(byteContent), &result)
+	return result.AccessToken, nil
+}
```
mikeland73, line 60: The auth library has a function that already provides the user and tokens, handles login/refresh, etc.

envsec/internal/auth/aws.go
```diff
@@ -0,0 +1,155 @@
...
+		Logins:     logins,
+	})
+	if err != nil {
+		return nil, err
+	}
+	debug.Log("aws credentials: %v \n\n", output.GoString())
+	return output, nil
+}
+
+func GetParameters() ([]*ssm.Parameter, error) {
```
mikeland73, line 122: I assume this is for testing?  If so, just add a comment indicating for testing.

FYI, `envsec.NewStore` takes a config with credentials. So once you have the cognito provided credentials you can initialize envsec and the rest should work.


Once we integrate, I think the general flow should be something like this:

[genConfig](https://github.com/jetpack-io/opensource/blob/main/envsec/internal/envcli/flags.go#L49-L49) will check if there's an authenticated user. If there is, it can pass the credentials and everything works. If there's no authenticated user and user passes flags we can assume user wants to use default credential chain (e.g. `.aws` or env variables). If there's no user and no flags we can show an error telling the user to log in _or_ pass the flags and ensure they have their own credentials.

mikeland73 (changes requested):

envsec/envsec.go
```diff
@@ -20,12 +22,20 @@ type EnvId struct {
...
+	projectId, err := jetcloud.ProjectID(wd)
+	if err != nil {
+		return EnvId{}, errors.New("Failed to retrieve projectID")
+	}
+	if projectId.String() == "" {
 		return EnvId{}, errors.New("ProjectId can not be empty")
 	}
 	return EnvId{
-		ProjectId: projectId,
+		ProjectId: projectId.String(),
```
mikeland73, line 38: I don't think we should make these changes to NewEnvId. They make this project not work as an open source project anymore.

envsec/internal/auth/user.go
```diff
@@ -66,11 +84,18 @@ func (u *User) ID() string {
...
-	if u == nil || u.IDToken == nil {
+func (u *User) GetAccessToken() string {
+	if u == nil || u.AccessToken == nil {
 		return ""
 	}
-	return u.IDToken.Claims.(jwt.MapClaims)["org_id"].(string)
+	return u.AccessToken.Raw
+}
+
+func (u *User) GetOrgId() string {
```
mikeland73, line 94: golang functions should not be prefixed with `Get` if possible.

envsec/internal/auth/user.go
```diff
@@ -66,11 +84,18 @@ func (u *User) ID() string {
 	return u.IDToken.Claims.(jwt.MapClaims)["sub"].(string)
 }
 
-func (u *User) OrgID() string {
-	if u == nil || u.IDToken == nil {
+func (u *User) GetAccessToken() string {
```
mikeland73, line 87: If you make the AccessToken public you don't need this function.

envsec/internal/envcli/flags.go
```diff
@@ -26,12 +26,12 @@ func (f *configFlags) register(cmd *cobra.Command) {
...
-		"org-id",
-		"",
-		"Organization id to namespace secrets by",
-	)
+	// cmd.PersistentFlags().StringVar(
+	// 	&f.orgId,
+	// 	"org-id",
+	// 	"",
+	// 	"Organization id to namespace secrets by",
+	// )
```
mikeland73, line 34: Can we make this change in a way that doesn't break existing open source functionality?

envsec/internal/envcli/ls.go
```diff
@@ -32,7 +32,7 @@ func listCmd() *cobra.Command {
 				return err
 			}
 			// Populate the valid Environments
-			envNames := []string{"DEV", "PROD", "STAGING"}
+			envNames := []string{"dev", "prod", "staging"}
```
mikeland73, line 35: I think this may be a non-backward compatible change

envsec/parameter_store.go
```diff
@@ -144,32 +144,6 @@ func (s *parameterStore) ListByPath(ctx context.Context, path string) ([]EnvVar,
 	return results, nil
 }
 
-func (s *parameterStore) listByTags(ctx context.Context, envId EnvId) ([]EnvVar, error) {
```
mikeland73, line 147: why is this being removed?

envsec/ssm_store.go
```diff
@@ -89,36 +89,6 @@ func (s *SSMStore) DeleteAll(ctx context.Context, envId EnvId, names []string) e
 	return s.store.deleteAll(ctx, envId, names)
 }
 
-func buildFilters(envId EnvId) []types.ParameterStringFilter {
```
mikeland73, line 92: Same here, why is this being removed?

envsec/util.go
```diff
@@ -11,11 +11,17 @@ import (
...
+
+func orgPath(envId EnvId) string {
+	// because of aws permissions orgPath needs to have a trailing `/`
+	return path.Join(PATH_PREFIX, envId.OrgId) + "/"
 }
 
 func projectPath(envId EnvId) string {
-	return path.Join(PATH_PREFIX, envId.ProjectId)
+	// because of aws permissions projectPath needs to have a trailing `/`
+	return path.Join(orgPath(envId), envId.ProjectId) + "/"
```
mikeland73, line 24: This will break backward compatibility.

mikeland73 (changes requested):
@mohsenari @loreto  let's sync up on this. 

I think there's a way to keep the general structure of envsec while providing federated credentials. The changes in this PR would essentially make envsec non-functional for a user that provodes own credentials.

mohsenari:
Closing this PR since #115 and #116 basically do the same thing in 2 separate PRs

PR #115, opened 2023-08-29
[envsec] Add awsfed library to fetch aws federated credentials

mikeland73:
## Summary

This is inspired by (copied from) https://github.com/jetpack-io/opensource/pull/112 (thanks @mohsenari !)

but makes a few changes:

* Use aws go sdk v2 (instead of v1)
* Remove `GetSSMConfig` function which creates a circular dependency (This can be included in envcli later on)
* Changes to auth package to simplify struct

## How was it tested?

Added `awsfed.New().AWSCreds(cmd.Context(), user.AccessToken)` and printed results


envsec/internal/auth/user.go
```diff
@@ -73,7 +74,7 @@ func (u *User) OrgID() string {
 	return u.IDToken.Claims.(jwt.MapClaims)["org_id"].(string)
```
mohsenari, line 74: Are we just reading the `org_id` claim or are we attempting to read `https://auth.jetpack.io/org_id` from the token as well?
If that's the case or if we want to read both, this needs to be updated to accommodate for it. 
Also, can we change all `org_id`'s to `org-id` to align with current implementation of envsec?

envsec/internal/auth/user.go
```diff
@@ -84,10 +85,17 @@ func (a *Authenticator) parseToken(stringToken string) (*jwt.Token, error) {
...
+		return nil, errors.WithStack(err)
+	}
+	idToken, err := jwt.Parse(tokens.IDToken, jwks.Keyfunc)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 
-	return token, nil
+	return &User{
+		AccessToken: accessToken,
```
mohsenari, line 98: To get a working accessToken as a string, you need to reference `User.AccessToken.Raw` because of that, can we make this a private field and make a method named `(u *User) AccessToken()` that returns `u.accessToken.Raw`?

mohsenari (approved):

envsec/internal/auth/user.go
mikeland73: This function is most likely going away (it's not needed). The org_id is only used by the server (the server should validate the JWT and extract it on it's own, that's the secure way to do it). I'll remove this after I make the server changes

envsec/internal/auth/user.go
mikeland73: I kinda prefer using *jwt.Token over strings. 

For example awsfed uses *jwt.Token param which is safer than using a string (strings are more common so there's less compile time type enforcement)

Similarly `jetcloud` currently uses `user` because it is using the temporary `OrgID()` method, but in reality it should use just a token.

mikeland73 (commented):
@mohsenari I'm gonna merge this to keep building on it, but we can definitely continue conversation about https://github.com/jetpack-io/opensource/pull/115/files#r1309555833 happy to change if you feel strongly and don't think my comment makes sense.

PR #117, opened 2023-08-30
[envsec] added caching for jwks public keys

mohsenari:
## Summary
a few implementation decisions:
- The caching is under `auth` package. 
- Cache duration is 1 hour. I check file's last modified. That means a user can manually modify the file to keep the cache going. But because it is client side any approach will have its flaws. So I rather discuss before implementing a solution for it.
- file name is `<domain>.jwks.json`
- readJWKSCache() checks file age as well as reading the cached file. Returns nil in case of cache miss.
- saveJWKSCache() downloads jwks.json and saves the values in a file and returns it. To prevent an extra read cache call after saving the cache.

## How was it tested?
- devbox run build
- envsec ls

### DIFF
--- a/envsec/internal/auth/cache.go
+++ b/envsec/internal/auth/cache.go
@@ -12,17 +12,19 @@ import (
 	"github.com/pkg/errors"
 )
 
-const dirName = ".jetpack"
+const dirName = "jetpack"
 
 func (a *Authenticator) fetchJWKSWithCache() (*keyfunc.JWKS, error) {
 	jwksURL := fmt.Sprintf("https://%s/.well-known/jwks.json", a.Domain)
-	cacheFileName := fmt.Sprintf("%s.jwks.json", a.Domain)
-	cacheBaseDir, err := os.UserCacheDir()
+	fileName := fmt.Sprintf("%s.jwks.json", a.Domain)
+	baseDir, err := os.UserCacheDir()
 	if err != nil {
-		cacheBaseDir = "~/.cache"
+		baseDir = "~/.cache"
 	}
-	// example ~/.cache/.jetpack/auth.jetpack.io.jwks.json
-	path := filepath.Join(cacheBaseDir, dirName, cacheFileName)
+	// example ~/.cache/jetpack/auth.jetpack.io.jwks.json
+	cacheDir := filepath.Join(baseDir, dirName)
+	path := filepath.Join(cacheDir, fileName)
+
 	// check Cache if miss, jwksJSON will be empty
 	jwksJSON, err := readJWKSCache(path)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


envsec/internal/auth/cache.go
```diff
@@ -0,0 +1,88 @@
...
+const jwksFileName = "jwks.json"
+const cacheDuration = 1 * time.Hour
+
+func (a *Authenticator) fetchJWKSWithCache() (*keyfunc.JWKS, error) {
+	jwksURL := fmt.Sprintf("https://%s/.well-known/jwks.json", a.Domain)
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	path := filepath.Join(wd, dirName, jwksFileName)
```
mikeland73, line 25: Use `os.UserCacheDir()` (similar to `getAuthFilePath`) that way multiple projects can share jwks.

The file should look like:

`filepath.Join(os.UserCacheDir(), "jwks", domain)` or something like that. That way all tools can share cache.

envsec/internal/auth/cache.go
```diff
@@ -0,0 +1,88 @@
...
+	"path/filepath"
+	"time"
+
+	"github.com/MicahParks/keyfunc/v2"
+	"github.com/pkg/errors"
+)
+
+const dirName = ".jetpack"
+const jwksFileName = "jwks.json"
+const cacheDuration = 1 * time.Hour
```
mikeland73, line 17: `1` is not needed

envsec/internal/auth/cache.go
```diff
@@ -0,0 +1,88 @@
...
+	fileInfo, err := os.Stat(path)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	modificationTime := fileInfo.ModTime()
+	current := time.Now()
+	if current.After(modificationTime.Add(cacheDuration)) {
```
mikeland73, line 56: I find this easier to read if you inline `time.Hour`

envsec/internal/auth/cache.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, nil
+	}
+	byteContent, err := os.ReadFile(path)
+	if err != nil {
+		return nil, nil
+	}
+	return byteContent, nil
```
mikeland73, line 63: `return os.ReadFile(path)`

mikeland73 (approved):
This is great. Only big comment is to store cache in a more global cache location.

envsec/internal/auth/cache.go
mohsenari: I went with 
```go
// example ~/.cache/.jetpack/auth.jetpack.io.jwks.json
path := filepath.Join(cacheBaseDir, dirName, cacheFileName)
```
cacheFileName includes the domain. Both auth0 and stytch idp domains can share the same directory but different file names.

PR #118, opened 2023-08-30
[envsec] fix ls envname bug

mohsenari:
## Summary
All envsec commands (except `ls` and `exec`) use `strings.ToLower()` when passing `EnvId.EnvName` that means environments like "DEV" and "PROD" becomes "dev" and "prod" in aws ssm.

This creates a bug where you can set a secret with `envsec set key=value` but in `envsec ls` not be able to see that key. This change might be backwards incompatible, but I don't know how envsec at its current state is usable unless someone specifies the environment everytime they `set` and `ls`

Screenshot 2023-08-30 at 6 53 31 PM

## How was it tested?
- `envsec set testkey=testvalue`
- `envsec ls`

### DIFF
--- a/envsec/internal/envcli/exec.go
+++ b/envsec/internal/envcli/exec.go
@@ -34,7 +34,7 @@ func execCmd() *cobra.Command {
 			commandToRun := exec.Command("/bin/sh", "-c", commandString)
 
 			// Default environment to fetch values from is DEV
-			envNames := []string{"DEV"}
+			envNames := []string{"dev"}
 			// If a specific environment was set by the user, then just use that one.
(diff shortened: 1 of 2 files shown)


ipince (approved):

mikeland73 (changes requested):
Can we hold off on the backward incompatible changes until we have a plan for schema?

mikeland73 (approved):
On second view, this won't break backwards compatibility because launchapad uses envsec as a library and not a CLI. 

This means that:

* envsec the library is case sensitive.
* envsec the CLI tool is case insensitive (and stores everything lowercase).

I think this is OK.

Could you also modify the hard coded envNames (in exec.go and ls.go) to be all lower case?

PR #119, opened 2023-08-30
[easy][envsec] Fix env case

mikeland73:
## Summary

Not sure we wanna keep uppercase, but for now this needs to match the default.

### DIFF
--- a/envsec/internal/envcli/flags.go
+++ b/envsec/internal/envcli/flags.go
@@ -40,7 +40,7 @@ func (f *configFlags) register(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(
 		&f.envName,
 		"environment",
-		"dev",
+		"DEV",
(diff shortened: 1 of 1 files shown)


mohsenari:
@mikeland73 please see #118 when we use `envsec set` and other commands we lowercase the env name.

ipince:
Also, FWIW, i prefer lowercase

mikeland73:
@mohsenari https://github.com/jetpack-io/opensource/pull/118 is backwards incompatible. The bug you ran into is that the default flag for environment is `dev` but the default for everything else is `DEV`. This fixes that issue, but it doesn't change the flag values for now.

@ipince I agree lower case is better. We're just working out how to do it while continuing to support launchpad.

mikeland73:
Closing in favor of https://github.com/jetpack-io/opensource/pull/118

PR #120, opened 2023-08-31
[easy][envsec] Don't send org-id (it is in token)

mikeland73:
## Summary

TSIA

### DIFF
--- a/envsec/internal/jetcloud/client.go
+++ b/envsec/internal/jetcloud/client.go
@@ -7,8 +7,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"os"
+	"path"
 
 	"go.jetpack.io/envsec/internal/auth"
 	"go.jetpack.io/envsec/internal/envvar"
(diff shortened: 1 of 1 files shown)


envsec/internal/jetcloud/client.go
```diff
@@ -53,8 +53,6 @@ func (c *client) newProjectID(ctx context.Context, user *auth.User, repo, subdir
 	fmt.Fprintf(os.Stderr, "Creating new project for repo=%s subdir=%s\n", repo, subdir)
 
 	p, err := post[project](ctx, c, user, map[string]string{
-		// TODO: org_id should be a claim in the ID token, not passed as post data
-		"org_id":   user.OrgID(),
```
mohsenari, line 57: This is ok. But I'm wondering how does the concept of organizations can work with users who use their own aws parameter store? Will the have no communication with jetpack for initializing a project and getting project id?

mohsenari (approved):

envsec/internal/jetcloud/client.go
mikeland73: This code path doesn't get used at all by non-jetpack cloud users. For open source users, org is just a tag and part of the env path.

The idea here is that we need the server to extract and validate the org_id. We can't trust the client. Even if the client tried to use a "bad" org ID when using the AWS SSM SDK it would not work because of our identity federation setup.

PR #121, opened 2023-08-31
[envsec] Switch to stytch

mikeland73:
## Summary

Switch envsec auth to auth provider

## How was it tested?

Untested.


mohsenari (approved):
Let's check if identity pool id stays the same after terraform changes are applied

LucilleH (approved):
🚢 🚢

PR #122, opened 2023-08-31
[envsec] Allow custom path

mikeland73:
## Summary

Allow custom var path and namespace. This can be used by launchpad to co continue backward compatibility.


mohsenari (approved):
Lgtm

PR #123, opened 2023-09-01
[envsec]  list by path

mohsenari:
## Summary
Added a check to see if user is authenticated with jetpack cloud, that means the parameter store keys follow `jetpack-data/env/org/project/<dev or prod>/name` convention. So we can reduce the number of aws API calls from 6 (3 envs each make a 1 descibeParam and 1 getParam call) in case of list by tags, to 1 call and use getparametersByPath.

This PR only focuses on `envsec ls` command but once it's merged, I can follow up with `exec` and `download` commands too.

## How was it tested?
Not fully tested yet


envsec/parameter_store.go
```diff
@@ -308,3 +317,9 @@ func awsSSMParamStoreValueToString(s *string) string {
 	}
 	return *s
 }
+
+func envNameAndKeyFromPath(s string) (string, string) {
```
mohsenari, line 321: returns `dev, testkey1` from `/jetpack-data/env/test_org_123/proj_01h8ynw132fdytr9hagq2xngp8/dev/testkey1`

envsec/envsec.go
```diff
@@ -37,6 +37,8 @@ func NewEnvId(projectId string, orgId string, envName string) (EnvId, error) {
 type Store interface {
 	// List all environmnent variables and their values associated with the given envId.
 	List(ctx context.Context, envId EnvId) ([]EnvVar, error)
+	// List all environmnent variables using parameter path.
+	ListByPath(ctx context.Context, envId EnvId, envNames []string) (map[string][]EnvVar, error)
```
mikeland73, line 41: Instead of adding a new function, `List` should default to `listByPath`

envsec/envsec.go
```diff
@@ -72,11 +74,11 @@ type Config interface {
...
 
 type SSMConfig struct {
-	Region          string
-	AccessKeyId     string
-	SecretAccessKey string
-	SessionToken    string
-	KmsKeyId        string
+	Region       string
+	AccessKeyId  string
+	SecretKey    string
```
mikeland73, line 79: AWS calls this `AWS_SECRET_ACCESS_KEY`. Any reason for change

mikeland73 (changes requested):
@mohsenari lets sync up on this. I would suggest a different approach:

* Envsec should be configurable so that the varPath function is modifiable via a constructor arg.  (This was done in https://github.com/jetpack-io/opensource/pull/122)
* The default implementation of envsec does not modify varPath function. It can listByPath **if possible** because it knows the structure.
* The **if possible** is because we need to ensure `envID` has enough parts to search by path. i.e. it must have all prefixes of path, so no empty spaces can exist in the path. If there's a gap in the path, we need to search by tags.
* If a custom varPath function is passed into `newSSMStore` then the store should fallback to listByTags.

This approach makes envsec independent from jetpack authentication and allows opensource users to take advantage of the performance optimizations of this approach. It also makes the library itself more efficient, instead of just the CLI.

PR #127, opened 2023-09-01
Add caching of tokens and support for ID claims

loreto:
## Summary
Change authentication library to save tokens, and to understand ID token claims. These changes are needed for the integration with the CLI.


sandbox/auth/internal/authflow/authflow.go
```diff
@@ -48,16 +49,21 @@ func (f *AuthFlow) OpenBrowser() error {
...
 		return nil, err
 	}
 
-	err = f.verify(tok)
+	tok, err := session.FromOauth2(otok)
 	if err != nil {
 		return nil, err
 	}
+
+	// err = f.verify(tok)
```
LucilleH, line 64: no verify?

sandbox/auth/session/idclaims.go
```diff
@@ -0,0 +1,63 @@
...
+	// Email scope:
+
+	Email         string `json:"email,omitempty"`
+	EmailVerified bool   `json:"email_verified,omitempty"`
+
+	// Profile scope:
+
+	Name       string           `json:"name,omitempty"`
+	GivenName  string           `json:"given_name,omitempty"`
+	FamilyName string           `json:"family_name,omitempty"`
```
LucilleH, line 32: not first name, last name?

sandbox/auth/session/idclaims.go
```diff
@@ -0,0 +1,63 @@
...
+	Username   string           `json:"username,omitempty"`
+	Website    string           `json:"website,omitempty"`
```
LucilleH, line 34: What's this? Personal website?

LucilleH (approved):

sandbox/auth/internal/authflow/authflow.go
loreto: I'm in the middle of re-implementing the token verification. I will re-enable it in a future change.

sandbox/auth/session/idclaims.go
loreto: The names come from the standard: https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims

sandbox/auth/session/idclaims.go
loreto: Yeah, the website of the user. Although, maybe I'll get rid of this and of username, since we're not really using them. What do you think?

loreto: Looking at the standard again, if I keep username, it should be`preferred_username`

sandbox/auth/session/idclaims.go
loreto: Should I keep only name for now?

LucilleH: I'm fine keeping them if it is part of the standard 🤷‍♀️

PR #131, opened 2023-09-07
[envsec] Use legacy token if needed

mikeland73:
## Summary

This change allows envsec to work end-to-end using auth0. This is ready for dogfooding with devbox if we want.

## How was it tested?

```
envsec init
envsec set FOO=BAR
envsec ls --show
```

### DIFF
--- a/envsec/internal/auth/user.go
+++ b/envsec/internal/auth/user.go
@@ -17,6 +17,8 @@ var orgIDClaim = envvar.Get(
 	"org_id",
 )
 
+const legacyOrgIDClaim = "https://auth.jetpack.io/org_id"
+
 var ErrNotLoggedIn = errors.New("not logged in")
 var ErrNoOrgID = errors.New("no org ID")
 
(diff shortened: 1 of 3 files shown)


envsec/internal/auth/user.go
```diff
@@ -17,6 +17,8 @@ var orgIDClaim = envvar.Get(
 	"org_id",
 )
 
+const legacyOrgIDClaim = "https://auth.jetpack.io/org_id"
```
LucilleH, line 20: oh, do we want the org id claim of the new token be also namespaced? `https://accounts.jetpack.io/org_id`? Or do we not care

LucilleH (approved):

envsec/internal/auth/user.go
mikeland73: We don't need to do it. auth0 [recommends/requires](https://auth0.com/docs/secure/tokens/json-web-tokens/create-custom-claims) it, but it's not strictly required by OIDC standard. It just makes the token larger and a bit harder to read.

mikeland73: I would leave the new one as `org_id` and I'll just clean everything up once we finish migrating.

PR #133, opened 2023-09-12
Use the new auth library from envsec

loreto:
## Summary
Replaces the authentication in the envsec cli by the new auth library,  allowing us to use Stytch tokens. Note that as written the code is pulling the auth library from `sandbox` which is *not* being published separately (so technically we're breaking the `envsec` repo if you try to build from there).

I thought this might be ok since we're not yet publishing envsec. And the next PR will publish the auth library and fix this. However, let me know if you'd like me to change the order of operations.

## How was it tested?
Ran locally


envsec/internal/jetcloud/client.go
```diff
@@ -90,6 +88,7 @@ func post[T any](ctx context.Context, c *client, user *auth.User, data any) (*T,
 		return nil, err
 	}
 
+	fmt.Printf("Response: %s\n", body)
```
mikeland73, line 91: rm

envsec/internal/jetcloud/project.go
```diff
@@ -40,12 +40,17 @@ func InitProject(ctx context.Context, user *auth.User, wd string) (typeids.Proje
...
+	projectID, err := newClient().newProjectID(ctx, tok, repoURL, subdir)
 	if err != nil {
 		return typeids.NilProjectID, err
 	}
 
-	orgID, err := user.OrgID()
+	claims := tok.IDClaims()
+	if claims == nil {
+		return typeids.NilProjectID, errors.Errorf("token did not contain an org")
+	}
```
mikeland73, line 51: This code will never get exercised because newProjectID will fail if required claims are not set. 

In fact, any token without claims or missing claims like orgID should not be a valid token and maybe rejected further up the stack?

mikeland73 (approved):

envsec/internal/jetcloud/project.go
loreto: Right now the token from the auth library allows for cases when you have an access token, but no ID token. If we want to always, 100% guarantee the existence of the ID token, I can change the struct so the claims are *not* a pointer (and thus, can't be null)

Should we go that route?

envsec/internal/jetcloud/client.go
loreto: Fixed

PR #137, opened 2023-09-15
[envsec] Add global --json-errors flag

mikeland73:
## Summary

This helps devbox understand envsec errors. Added a global error flag instead of individual `--format` flags because it is easier to do and not all commands will have `--format`.

## How was it tested?
`./dist/envsec init --json-errors`

### DIFF
--- a/envsec/cmd/envsec/main.go
+++ b/envsec/cmd/envsec/main.go
@@ -5,10 +5,11 @@ package main
 
 import (
 	"context"
+	"os"
 
 	"go.jetpack.io/envsec/internal/envcli"
 )
 
 func main() {
-	envcli.Execute(context.Background())
+	os.Exit(envcli.Execute(context.Background()))
 }
--- a/envsec/internal/envcli/root.go
+++ b/envsec/internal/envcli/root.go
@@ -5,12 +5,20 @@ package envcli
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/spf13/cobra"
 )
 
-func EnvCmd() *cobra.Command {
+type rootCmdFlags struct {
+	jsonErrors bool
+}
+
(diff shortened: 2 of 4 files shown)


LucilleH (approved):
I have 2 questions:
1. We have `--json` flag as well. Should that cover the case for json errors?
2. In what scenario does devbox need to understand envsec errors? Can devbox simply prints envsec errors and fail?

mikeland73:
> We have --json flag as well. Should that cover the case for json errors?

Since not all commands have `--format` and it would be a bit challenging to maintain that, I thought doing errors separately with a global flag would scale better. It ensures all commands can do it. An alternative would be only output `json` errors if the command supports it but the downside is some commands don't have any usable output at all and would still require adding the flag. I'm open to both approaches, just thought this way would be easier to maintain.

> In what scenario does devbox need to understand envsec errors? Can devbox simply prints envsec errors and fail?

Cases I've seen so far:
* Not logged in (Hoping to fix this by passing token instead)
* Project doesn't belong to organization
* Directory is already initialized (will fix by adding flag to ignore error) 

Commands can fail for many reasons (known and unknown). I thought having structured data (and metadata if needed) would make the integration nicer. If we understand what went wrong we might also be able to do something about it.

LucilleH:
> * Not logged in (Hoping to fix this by passing token instead)

Wait, our login is not unified? envsec is not reading the same token as devbox?

> * Directory is already initialized (will fix by adding flag to ignore error)

So, `envsec init` errors when the directory is already initialized? Should it just return ok?

> * If we understand what went wrong we might also be able to do something about it

That means a custom integration between devbox and envsec, which I thought we want to avoid initially

mikeland73:
> Wait, our login is not unified? envsec is not reading the same token as devbox?

We are actually going to share tokens but in the future we may not (because different apps should have different client IDs). For simplicity we will reuse the client ID.

> So, envsec init errors when the directory is already initialized? Should it just return ok?

Yeah, right now it returns exit code 1, but it should probably be 0. Will fix.

PR #138, opened 2023-09-15
[envsec] Add ability to specify id-token via flag

mikeland73:
## Summary

Stacked on https://github.com/jetpack-io/opensource/pull/137

Add `--id-token` flag to envsec.

Having devbox and envsec assume the same location for tokens feels brittle. Allowing it to be passed in as a flag makes it more robust.

Also I tried to make a few improvements to auth package. Added some comments and tried to reduce redundant data in the struct. @loreto let me know what you think.

## How was it tested?
Tested using auth0 token (because `envsec auth login` is not currently working for me.

```bash
./dist/envsec auth whoami --id-token <token>
Logged in
User ID: ...
Email: mike...
Name: Mike...
```

### DIFF
--- a/sandbox/auth/session/token.go
+++ b/sandbox/auth/session/token.go
@@ -2,16 +2,40 @@ package session
 
 import (
 	"github.com/go-jose/go-jose/v3"
+	"github.com/go-jose/go-jose/v3/jwt"
 	"golang.org/x/oauth2"
 )
 
+// Tokens can generally be ID or access. It's a bit weird to have a struct
+// named "Token" that only contains an ID token. An alternative is to just rename
+// this struct to IDToken.
 type Token struct {
-	oauth2.Token // Embed and oauth2 token
-
+	// Are these fields needed?
 	Issuer string            `json:"issuer,omitempty"`
 	Keys   []jose.JSONWebKey `json:"keys,omitempty"`
 
 	IDToken string `json:"id_token,omitempty"`
 
 	idClaims *IDClaims
+}
+
+func TokenFromString(t string) (*Token, error) {
+	_, err := jwt.ParseSigned(t)
+	if err != nil {
+		return nil, err
+	}
+	tok := &Token{
+		IDToken: t,
+	}
+	claims := tok.IDClaims()
+	if claims != nil {
+		tok.Issuer = claims.Issuer
+	}
+	return tok, nil
+}
+
+func (t *Token) ToOauth2Token() *oauth2.Token {
+	return &oauth2.Token{
+		AccessToken: t.IDToken,
+	}
 }
--- a/envsec/cmd/envsec/main.go
+++ b/envsec/cmd/envsec/main.go
@@ -5,10 +5,11 @@ package main
 
 import (
 	"context"
+	"os"
 
 	"go.jetpack.io/envsec/internal/envcli"
 )
 
 func main() {
-	envcli.Execute(context.Background())
+	os.Exit(envcli.Execute(context.Background()))
 }
--- a/envsec/internal/envcli/auth.go
+++ b/envsec/internal/envcli/auth.go
@@ -6,11 +6,14 @@ package envcli
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/auth"
 	"go.jetpack.io/auth/session"
 	"go.jetpack.io/envsec/internal/envvar"
 )
+
+var errNotLoggedIn = errors.New("not logged in")
 
 func authCmd() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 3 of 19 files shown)


sandbox/auth/session/token.go
```diff
@@ -2,16 +2,40 @@ package session
 
 import (
 	"github.com/go-jose/go-jose/v3"
+	"github.com/go-jose/go-jose/v3/jwt"
 	"golang.org/x/oauth2"
 )
 
+// Tokens can generally be ID or access. It's a bit weird to have a struct
```
loreto, line 9: So, this struct is actually meant to represent both the id token and the access token (the access token was part of the oauth2.Token field which you removed).

It's the equivalent of what you used to call `TokenSet` in the previous library; I also considered naming it `Session`.

I ended up naming it `Token` just to match the nomenclature of `oauth2.Token`, but if you think that name is confusing, I'd be happy to go with `TokenSet` or `Session` instead.

sandbox/auth/session/token.go
```diff
@@ -2,16 +2,40 @@ package session
...
+// named "Token" that only contains an ID token. An alternative is to just rename
+// this struct to IDToken.
 type Token struct {
-	oauth2.Token // Embed and oauth2 token
```
loreto, line 9: I don't think we should remove this (although I'm open to hearing why you think we should). The idea was to "extend" the standard Oauth token.

This embedded struct does a couple of things:
1. Ensures all the appropriate oauth2.Token fields get serialized when we convert to JSON (for example, the Access Token is getting serialized because of this field, but no longer would be once you remove)

2. It lets us pass our token to places that need an `oauth2` token without a conversion.

3. It gives us access to methods we don't have to re-implement. For example, I was going to use `.Valid()` (https://pkg.go.dev/golang.org/x/oauth2#Token.Valid) to address this bug: "New auth lib returns ID token even if expired".

sandbox/auth/session/token.go
```diff
@@ -2,16 +2,40 @@ package session
...
-
+	// Are these fields needed?
```
loreto, line 13: Issuer and Keys are needed to add validation when deserializing the token. I haven't implemented that, but was hoping to add at some point.

sandbox/auth/session/token.go
```diff
@@ -2,16 +2,40 @@ package session
...
+	}
+	claims := tok.IDClaims()
+	if claims != nil {
+		tok.Issuer = claims.Issuer
+	}
+	return tok, nil
+}
+
+func (t *Token) ToOauth2Token() *oauth2.Token {
+	return &oauth2.Token{
```
loreto, line 38: If we end up agreeing on leaving the `oauth2.Token` field alone, then we can remove this function.

If we do keep it, I have a worry around losing information. The original `oauth2.Token` has the following fields that we'd be losing: `TokenType`, `RefreshToken`, `Expiry`

In particular, I believe that `RefreshToken` and `Expiry` is what make it possible for the underlying oauth library to automatically do token refreshes. I believe that with this code we don't get any refreshing anymore.

loreto (commented):
Added comments regarding the auth library changes.

Re: id-token flag, not sure I follow how this interacts with the devbox integration. Could we talk live about that so that I understand?

sandbox/auth/session/token.go
mikeland73: Issuer should be a claim in the token and keys should be reachable via jwks endpoint no?

sandbox/auth/session/token.go
mikeland73: I'm a bit confused because in line 12 in sandbox/auth/session/oauth.go you are passing the ID token into this field, so we essentially have it twice.

The `AccessToken` field inside is a bit unfortunatel becase it doesn't look like anything in this token is specifically access vs ID.

We can keep this, but then we should remove the raw string IDtoken. Or if this is meant to be the access token then we should rename the struct to tokenSet and the fields accordingly.

sandbox/auth/session/token.go
loreto: Yes, but then you need to:
Decode Token without verifying the signature -> Do a network roundtrip using OIDC discovery to figure out the right endpoints -> Do another network roundtrip to download the keys -> Verify signature

Vs:
Decode and verify token with the "cached" issuer and JWKS data immediately (no network). I know there's an edge case where, if the keys have been rotated, you still need to do another network request to get the latest keys; but that can be done only if the cached data fails to verify the token.

sandbox/auth/session/token.go
mikeland73: You are passing the ID token into both fields. Did you mean to do that? 

TokenSet or Session both make sense to me, but `Token` feels wrong if there are actually multiple tokens (id, access, refresh)

It is weird that `oauth2.Token` has access token and refresh so we need to use that in a way that makes sense for us.

A typical setup is to save a bundle that contains access_token, id_token, refresh_token, and maybe an expiration for the refresh (because refresh tokens tend to not be JWTs). The reason all of these are bundled together is because auth providers will return all of them together as part of the auth flow.

sandbox/auth/session/token.go
mikeland73: Good point, I messed this up. Agree we should not lose that info.

PR #144, opened 2023-09-18
stackerr: annotate errors with source file and line number

gcurtis:
I wasn't entirely sure which directory this should go under. Let me know if it should be moved.

A more basic version of pkg/errors that offers a `fmt.Errorf` function which can record the filename and line number of where it was called.

Example:

	user := "gcurtis"
	wrapped := stackerr.Errorf("wrong password")
	err := stackerr.Errorf("login %q: %w", user, wrapped)
	fmt.Printf("error: %+v\n", err)

	// Output:
	// error: login "gcurtis": wrong password
	// /go/src/handler.go:176 login "gcurtis": wrong password
	// /go/src/login.go:14 wrong password

### DIFF
--- a/pkg/stackerr/example_test.go
+++ b/pkg/stackerr/example_test.go
@@ -1,0 +1,42 @@
+package stackerr
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Examples are in their own file because their output has hard-coded line
+// numbers. This makes it easier to change/add other tests without constantly
+// breaking the examples.
+
+func Example_wrapped() {
+	// user := "gcurtis"
+	wrapped := Errorf("wrong password")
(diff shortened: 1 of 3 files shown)


mikeland73 (approved):
Nice! I think later on we may want full stack for logging but this is definitely good enough to replace `pkg/errors`

In terms of where to put new projects:

* /pkg/sandbox -> just messing around, but want to commit my code
* /pkg -> ready to be shared with other projects internally
* / -> Want to open source as own project.

loreto:
Is this PR still live? Or abandoned?

gcurtis:
@loreto I'd still like to merge it. I'll set aside a few minutes to make the last couple of changes on Monday.

PR #145, opened 2023-09-19
[code-quality] Turn on most linters for envsec

mikeland73:
## Summary

TSIA

## How was it tested?

`devbox run lint`

### DIFF
--- a/envsec/.golangci.yml
+++ b/envsec/.golangci.yml
@@ -19,19 +19,16 @@ linters:
     - reassign
     - revive
     - staticcheck
-    # - stylecheck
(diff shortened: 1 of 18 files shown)


loreto (approved):
LGTM

Is it easy enough to setup `.golangci.yml` at the root of the repo so that it applies to everything under `opensource`?

mikeland73:
> Is it easy enough to setup .golangci.yml at the root of the repo so that it applies to everything under opensource
 
It is, but not all code in the repo is compliant (only envsec) so we need to fix issues. I'll do that in follow up.

PR #152, opened 2023-09-20
[cicd] Enable devbox cache and add golang cache

mikeland73:
## Summary

* Enable devbox cache
* Add golang cache (maybe eventually devbox can handle this as well)

## How was it tested?

CICD

image

### DIFF
--- a/.github/workflows/monorepo-go.yml
+++ b/.github/workflows/monorepo-go.yml
@@ -21,6 +21,16 @@ jobs:
       - name: Install devbox
         uses: jetpack-io/devbox-install-action@v0.7.0
 
+      # can this be part of devbox action?
+      - name: Mount golang cache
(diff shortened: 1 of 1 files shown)


.github/workflows/monorepo-go.yml
```diff
@@ -20,6 +20,18 @@ jobs:
 
       - name: Install devbox
         uses: jetpack-io/devbox-install-action@v0.7.0
+        with:
+          enable-cache: true
+
+      # can this be part of devbox action?
```
LucilleH, line 26: no, I don't think so

LucilleH (approved):

.github/workflows/monorepo-go.yml
mikeland73: :(

PR #153, opened 2023-09-20
[auth] Set success and failure redirect by flag

LucilleH:
## Summary
Set success and failure redirect by flag. Default to the jetpack urls

## How was it tested?
```
go build -o dist/auth cmd/auth/main.go
./dist/auth login https://accounts.jetpack.io --client <client-id> --success https://www.jetpack.io/account/login/success?product=envsec
```


mikeland73 (commented):
Why put this into the client? We could add a `redirect` query param to the callback server so the auth flow itself controls the redirect.

My main concern is hardcoding URLs that may change into the binary. We also can't control the redirect depending on the user and context.

LucilleH:
> Why put this into the client? We could add a `redirect` query param to the callback server so the auth flow itself controls the redirect.
> 
> My main concern is hardcoding URLs that may change into the binary. We also can't control the redirect depending on the user and context.

Good point. Closing.

PR #154, opened 2023-09-20
[runx+envsec] Use same cache location as auth pkg

mikeland73:
## Summary

Use consistent cache locations everywhere.

Fix envsec aws credentials cache bug where cache key did not take into account changing organizations.


loreto (approved):
LGTM

1. What's your thought on whether we should make the directory `go.jetpack.io` vs simply `jetpack.io` for everything?

2. Could we use SHA256 instead of MD5 for the hash. Most modern stores are deprecating MD5 (docker uses SHA256, git uses MD5 but introduced support for SHA256 with the hope of migrating to it over time)

3. Thought for later: would it make sense for awsfed to become part of the auth library, and that library can then handle caching for all the tokens (without having to store them in an envsec specific place)

mikeland73:
> What's your thought on whether we should make the directory go.jetpack.io vs simply jetpack.io for everything
I'm good either way as long as we're consistent. I guess `jetpack.io` is a bit better because it is less golang-ish

> Could we use SHA256 instead of MD5 f
will change in this PR

> Thought for later: would it make sense for awsfed to become part of the auth library
agree

envsec/internal/filecache/filecache.go
```diff
@@ -68,7 +65,11 @@ func (c *cache) Get(key string) ([]byte, error) {
...
 func (c *cache) filename(key string) string {
-	dir := xdg.CacheSubpath(c.appName)
+	cacheDir, err := os.UserCacheDir()
+	if err != nil {
+		cacheDir = "~/.cache"
+	}
+	dir := filepath.Join(cacheDir, c.domain)
 	_ = os.MkdirAll(dir, 0755)
-	return xdg.CacheSubpath(filepath.Join(c.appName, prefix+key))
+	return filepath.Join(dir, key)
```
LucilleH, line 74: Re: CICD failure: Could it be this line `os.MkdirAll(dir, 0755)` creating the `~/.cache` dir under a specific user?

PR #157, opened 2023-09-21
[envsec] Bump pkg version to ensure auth location is correct

mikeland73:
## Summary

This is an unfortunate needed step. When we make changes in pkg, downstream dependencies don't get the newest pkg. This means we need 2 commits. We could automate this, but it would still require an additional commit.

## How was it tested?

builds


loreto (approved):
LGTM

What command do we use to bump? Wondering if we can make it part of `devbox run go-mod-sync`

mikeland73:
> What command do we use to bump? Wondering if we can make it part of devbox run go-mod-sync

We can't bump in the same PR. We need to publish to repo first.

PR #158, opened 2023-09-21
[auth] change GetSession so it returns ok

mikeland73:
## Summary

This will make code more readable and reduce mistakes when using library. An invalid token is not super useful to client.

Updated error messages for new interface.

I still think an error is better (can help with refresh failures, etc) but @loreto you mentioned preferring this. 

## How was it tested?

Untested.


loreto (approved):
LGTM

PR #160, opened 2023-09-21
[envsec] Don't allow init outside of git repo

mikeland73:
## Summary

Prints error if envsec init is used outside of repo. Updates dependencies.

## How was it tested?

`envsec init` in and outside of repo.


loreto (approved):
LGTM

loreto:
FYI There's a go.mod to fix

PR #166, opened 2023-10-02
[envsec] Ensure id token is refreshed

mikeland73:
## Summary

* Ensure refresh updates id token, previously it was only updating access token.
* Changed GetSession and a few internal functions to return errors instead of just swallowing it up. 

## How was it tested?

```
envsec auth logout
envsec auth whoami
envsec auth login
envsec auth whoami
envsec auth refresh
# inspected id token to ensure it was refreshed
```


loreto (approved):
LGTM

PR #170, opened 2023-10-04
[awsfed] Added ssm config from id token

mohsenari:
## Summary
Added a function to generate ssm config from id token as an input. The other genSSMConfig function was relying on the token from a cached file. In a web server setting, each user has their own idToken therefore, saving and caching each request's id token is not ideal.

### DIFF
--- a/envsec/pkg/awsfed/awsfed.go
+++ b/envsec/pkg/awsfed/awsfed.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/service/cognitoidentity"
 	"github.com/aws/aws-sdk-go-v2/service/cognitoidentity/types"
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/pkg/errors"
 	"go.jetpack.io/envsec"
 	"go.jetpack.io/envsec/internal/envvar"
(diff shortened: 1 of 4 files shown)


envsec/pkg/awsfed/awsfed.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/service/cognitoidentity"
 	"github.com/aws/aws-sdk-go-v2/service/cognitoidentity/types"
+	"github.com/golang-jwt/jwt/v5"
```
mikeland73, line 11: Use session.Token instead like existing functions.

Side note, could you make sure we're using `github.com/golang-jwt/jwt/v5` everywhere. I think we accidentally added v4s https://github.com/jetpack-io/axiom/blob/d4079463c5582b9906f5b8ba7b89ce39ee888218/pkg/jetdb/token.go#L6-L6

envsec/pkg/awsfed/awsfed.go
```diff
@@ -99,6 +100,48 @@ func (a *AWSFed) AWSCreds(
 	return output.Credentials, nil
 }
 
+// AWSCredsFromIdToken behaves similar to AWSCreds but it takes JWT from input
+// rather than reading from a file or cache. This is to allow web services use
+// this package without having to write every user's JWT in a cache or a file.
+func (a *AWSFed) AWSCredsFromIdToken(
```
mikeland73, line 106: This is copy pasting much of `AWSCreds`. You can replace the middle portion of that function.

Also suggested naming:

```
AWSCreds
AWSCredsWithLocalCache
```

envsec/pkg/awsfed/awsfed.go
```diff
@@ -129,3 +172,23 @@ func GenSSMConfigForUser(
 		Region:          fed.Region,
 	}, nil
 }
+
+func GenSSMConfigFromIdToken(
```
mikeland73, line 176: Maybe merge both functions and call them `GenSSMConfigFromToken` 

and add a `useCache` param.

envsec/pkg/awsfed/awsfed.go
mikeland73: Also I think the can be lower case? Not sure anything outside this package is using them

mikeland73 (changes requested):

mikeland73 (approved):

PR #171, opened 2023-10-05
[runx] Add support for github token, implement resolve

mikeland73:
## Summary

This PR does a few things:

* Adds support for github token. This prevents rate-limiting when getting `releases.json`
* Adds resolve to API. This will allow devbox to validate packages exist and resolve versions in order to use lockfile. 

@gcurtis I'm copying the devbox structure for the public interface (RunX interface) but was curious if you would do it differently. Specifically, we could remove the public interface and rename `impl` to just be `runx` and use a struct:

```golang
runx.New(runx.Opts{
  GithubAPIToken: "",
}).Install(...)
```

thoughts?

## How was it tested?

```
RUNX_GITHUB_API_TOKEN=<token> runx +jetpack-io/envsec envsec
RUNX_GITHUB_API_TOKEN=<token> pkg resolve jetpack-io/envsec@latest
RUNX_GITHUB_API_TOKEN=<token> pkg resolve jetpack-io/envsec@v0.0.3
```


pkg/sandbox/runx/impl/runx.go
```diff
@@ -0,0 +1,13 @@
+package impl
+
+import "go.jetpack.io/pkg/sandbox/runx/impl/runopt"
+
+type RunX struct {
+	GithubAPIToken string
+}
```
gcurtis, line 7: My preference would be to put this struct directly in the `runx` package and remove the interface. Because the GitHub API token is optional, I'd also remove `runopt` in favor of setting the `GithubAPIToken` field directly. For example:

```go
var r runx.Runx
// -- or --
// r := runx.RunX {
//     GithubAPIToken: "token",
// }
r.Run(ctx, "golangci-lint", "run")
```

pkg/sandbox/runx/impl/registry/registry.go
```diff
@@ -2,28 +2,38 @@ package registry
...
 	gh       *github.Client
 }
 
-func NewLocalRegistry(rootDir string) (*Registry, error) {
-	rootPath := fileutil.Path(rootDir)
-	err := rootPath.EnsureDir()
+func NewLocalRegistry(ctx context.Context, githubAPIToken string) (*Registry, error) {
+	cacheDir, err := os.UserCacheDir()
 	if err != nil {
+		cacheDir = "~/.cache"
```
gcurtis, line 24: The `~` won't expand if the path is used outside a shell. You probably want:

```suggestion
		cacheDir = os.Expandenv("$HOME/.cache")
```

gcurtis (approved):
Looks good to me! I think we should eschew the interface and just use a struct. The impl pattern in Devbox feels a bit awkward and led to a giant interface.

pkg/sandbox/runx/impl/registry/registry.go
mikeland73: ` os.UserCacheDir()` usually only errors if $HOME is empty. Will return error instead .

PR #172, opened 2023-10-09
[runx] Allow non-archived binaries

mikeland73:
## Summary

Some repos (e.g. https://github.com/mvdan/gofumpt) release files as binaries instead of archives. This expands runx to treat unknown files (that still match the platform and architecture name structure) as binaries. It only does so if it finds a compatible artifact but the type is unknown.

We could additionally test the mime type of the file, but it doesn't seem like binaries use consistent mime types.

## How was it tested?

`./dist/runx +mvdan/gofumpt gofumpt --help`

### DIFF
--- a/pkg/sandbox/runx/impl/registry/extract.go
+++ b/pkg/sandbox/runx/impl/registry/extract.go
@@ -56,8 +56,8 @@ func contentDir(path string) string {
 	return filepath.Join(path, contents[0].Name())
 }
 
-func createSymbolicLink(src, dest, repoName string) error {
-	if err := fileutil.EnsureDir(dest); err != nil {
+func createSymbolicLink(src, dst, repoName string) error {
+	if err := os.MkdirAll(dst, 0700); err != nil {
 		return err
 	}
 	if err := os.Chmod(src, 0755); err != nil {
@@ -69,8 +69,9 @@ func createSymbolicLink(src, dest, repoName string) error {
 	if strings.Contains(binaryName, repoName) {
 		binaryName = repoName
 	}
-	err := os.Symlink(src, filepath.Join(dest, binaryName))
+	err := os.Symlink(src, filepath.Join(dst, binaryName))
 	if errors.Is(err, os.ErrExist) {
+		// TODO: verify symlink points to the right place
 		return nil
 	}
 	return err
--- a/pkg/sandbox/runx/impl/registry/extract.go
+++ b/pkg/sandbox/runx/impl/registry/extract.go
@@ -2,8 +2,10 @@ package registry
 
 import (
 	"context"
+	"errors"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/codeclysm/extract"
 	"go.jetpack.io/pkg/sandbox/runx/impl/fileutil"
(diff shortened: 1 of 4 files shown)


pkg/sandbox/runx/impl/registry/extract.go
```diff
@@ -53,3 +55,23 @@ func contentDir(path string) string {
 	}
 	return filepath.Join(path, contents[0].Name())
 }
+
+func createSymbolicLink(src, dest, repoName string) error {
+	if err := fileutil.EnsureDir(dest); err != nil {
```
gcurtis, line 60: This could just be `os.MkdirAll`.

pkg/sandbox/runx/impl/registry/extract.go
```diff
@@ -53,3 +55,23 @@ func contentDir(path string) string {
...
+	binaryName := filepath.Base(src)
+	// This is a good guess for the binary name. In the future we could allow
+	// user to customize.
+	if strings.Contains(binaryName, repoName) {
+		binaryName = repoName
+	}
+	err := os.Symlink(src, filepath.Join(dest, binaryName))
+	if errors.Is(err, os.ErrExist) {
+		return nil
+	}
```
gcurtis, line 75: Might want to check that the symlink is pointing to the right binary if it already exists in case the user ever installs a different version.

pkg/sandbox/runx/impl/registry/registry.go
```diff
@@ -118,7 +118,12 @@ func (r *Registry) GetPackage(ctx context.Context, ref types.PkgRef, platform ty
 		return "", err
 	}
 
-	err = Extract(ctx, artifactPath, installPath.String())
+	if isKnownArchive(filepath.Base(artifactPath)) {
+		err = Extract(ctx, artifactPath, installPath.String())
+	} else {
+		// If we can't extract, treat as binary
```
gcurtis, line 124: Checking the first 4 bytes would be a pretty good heuristic for determining if it's meant to be executable:

- `!#` = shell script
- `0x7f454c46` = ELF (Linux)
- `0xfeedface` = macho 32 bit binary (macOS)
- `0xfeedfacf` = macho 64 bit binary (macOS)
- `0xcafebabe` = macho universal binary (macOS)

gcurtis (approved):

pkg/sandbox/runx/impl/registry/extract.go
mikeland73: The symlink is already in a versioned directory: (Both the symlink and the source)

```
ls -al ~/Library/Caches/jetpack.io/pkgs/mvdan/gofumpt/v0.5.0/darwin/arm64/       
gofumpt -> /Users/mike/Library/Caches/jetpack.io/pkgs/mvdan/gofumpt/v0.5.0/gofumpt_v0.5.0_darwin_arm64
```

So nice to have in case we ever change the download logic but not urgent. Will add TODO

pkg/sandbox/runx/impl/registry/registry.go
mikeland73: Looks like I also need to add little endian for mac binaries

PR #173, opened 2023-10-10
[envsec] Move envcli to envsec/pkg

mikeland73:
## Summary

This is needed by launchpad


mohsenari (approved):
Thanks for looking into this

PR #180, opened 2023-10-31
[auth] Adds auth hooks to fetch STS access token

mikeland73:
## Summary

Adds post-login and post-refresh hooks to auth library so we can fetch and replace access token with our a custom short term token.

## How was it tested?

Untested, but will test using new token endpoint

### DIFF
--- a/envsec/pkg/envcli/auth.go
+++ b/envsec/pkg/envcli/auth.go
@@ -4,11 +4,14 @@
 package envcli
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/spf13/cobra"
 	"go.jetpack.io/envsec/internal/envvar"
+	"go.jetpack.io/envsec/internal/jetcloud"
 	"go.jetpack.io/pkg/sandbox/auth"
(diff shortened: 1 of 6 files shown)


envsec/pkg/envcli/auth.go
```diff
@@ -131,5 +134,22 @@ func newAuthClient() (*auth.Client, error) {
 	// TODO: Consider making scopes and audience configurable:
 	// "ENVSEC_AUTH_SCOPE" = "openid offline_access email profile"
 	// "ENVSEC_AUTH_AUDIENCE" = "https://api.jetpack.io",
-	return auth.NewClient(issuer, clientID)
+	return auth.NewClient(
+		issuer,
+		clientID,
+		getShortTermAccessToken,
+		getShortTermAccessToken,
```
LucilleH, line 141: Why is this variable get passed twice?

LucilleH: Oh, one for post login, one for post refresh I see

LucilleH (approved):

PR #181, opened 2023-11-02
feat: added simple variant of typeid that uses text type

spa5k:
## Summary
This creates a simple variant of typeid sql generation that generates text type, rather than a combined type of varchar + uuid.

The reason it was done because a few database tools like SQLC were failing to correctly recognize the type from database during generation, plus it will make it easier for user to just copy the id's from the table and won't have to parse them again.

## How was it tested?
I added this in Postgres Docker, and tested it there.


spa5k:
@mikeland73 @savil @gcurtis Please review it, I don't see anyway to request review.

loreto:
@spa5k Thank you so much for contributing (and sorry for the delay, I missed this a few days ago).

I've added a bunch of comments inline. My biggest observation is this: the tuple implementation is very thoroughly [tested](https://github.com/jetpack-io/opensource/tree/main/typeid/typeid-sql/supabase). This `text` version is not, and I think there's several edge cases (particularly against the empty prefix) that are not being handle.

I see two paths forward:
1. Implement these using the tuple versions internally. It's a suggestion I make inline already. I think this route would make the `_txt` implementations pretty trivial, because we'd be relying in the tested tuple versions.

2. Add tests for the text version if you want to keep the implementations separate.

What are your thoughts?

spa5k:
@loreto I don't see any comments, I think you've forgotten to submit them

spa5k:
@loreto Hey I've updated the PR, decided to no longer create a new file, but used the existing print function to handle the cases, and added 2 new functions for the text case, and their tests too.

loreto (approved):
Thanks for doing this! LGTM

PR #183, opened 2023-11-03
[envsec] Show good error message when we fail to get access token

mikeland73:
## Summary

This change does the following:

* Show better error message if the server returns one. In particular, the server will return instructions with a link to user can activate plan.
* If we fail to get an access token from our STS service, still write id token and refresh toke to store but make it expired. This will ensure we have a valid refresh token and that the next time the user tries to use envsec (presumably after clicking on link and selecting a plan) it will refetch all tokens.

## How was it tested?

`ENVSEC_API_HOST=http://localhost:8080 ./dist/envsec auth whoami`

### DIFF
--- a/envsec/internal/jetcloud/client.go
+++ b/envsec/internal/jetcloud/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


mikeland73:
@loreto @mohsenari PTAL

mohsenari (approved):

PR #186, opened 2023-11-09
[auth] Move jetauth logic to /pkg

mikeland73:
## Summary

Move shared `jetauth` logic to `/pkg` so it can be used by devbox

## How was it tested?

lints and builds


loreto (approved):
Only thought is whether this should live in a subdir of `/pkg` specifically for things that are jetpack specific and that we never plan to "promote" to a standalone generic package.

mikeland73:
> Only thought is whether this should live in a subdir of /pkg specifically for things that are jetpack specific and that we never plan to "promote" to a standalone generic package.

SGTM

mikeland73:
@loreto ugh sorry, I made the changes you requested, forgot to push and merged anyway. Will do in follow up

PR #187, opened 2023-11-10
[typeid] Unify typed and untyped versions

loreto:
## Summary
Unifies typed and untyped versions of typeid. Basically we now have a single version of the library that uses generics.
To support the "untyped" case where you might be dealing with an arbitrary (and unknown) prefix from a string, we have
a special `AnyPrefix` type.

A follow up PR will: make `Nil` equal to the empty struct. I also think I can incorporate Mike's changes so that they are optional (that way the user can choose whether they want to use reflection or not).

Note that in order to fully support the "typed" and "untyped" cases, I often need to create two versions of a method:
+ The typed version *only* takes a Prefix Type, and it "infers" the prefix string from the type.
+ The untyped version takes the prefix as a string, and fixes the type of the typeid to `TypeID[AnyPrefix]`

The main thing I'd like feedback on is whether we like this dual method approach, and whether we like the names I've picked. As an example, we used to have:
```
typeid.New("prefix")
```

But now there's:
```
typeid.New[PrefixType]()
// and
typeid.FromPrefix("prefix")
```

Do we like that? Do we prefer to keep the name `New("prefix")` for the "untyped" version, and introduce a new `NewTyped[Prefix]()` version?

Similar questions happen between, say `FromUUID` and the new `FromUUIDPrefix`.

## How was it tested?
All tests still pass.


typeid/typeid-go/typeid.go
```diff
@@ -10,44 +10,49 @@ import (
 )
 
 // TypeID is a unique identifier with a given type as defined by the TypeID spec
-type TypeID struct {
-	prefix string
+type TypeID[P Prefix] struct {
+	prefix P
```
mikeland73, line 14: This structure wont be compatible with an interface where `New` returns type `T`.

Specifically, this change would make the approach in this [PR](https://github.com/jetpack-io/opensource/pull/185) undoable. You could not be able to do:

`id, err := typeid.New[MyID]()`

typeid/typeid-go/typeid.go
```diff
@@ -10,44 +10,49 @@ import (
...
+	}
 }
 
-// New returns a new TypeID with the given prefix and a random suffix.
-// If you want to create an id without a prefix, pass an empty string.
-func New(prefix string) (TypeID, error) {
-	return From(prefix, "")
+// New returns a new TypeID of the given type.
+// The prefix is inferred from the type, and the suffix is random.
+func New[P Prefix]() (TypeID[P], error) {
```
mikeland73, line 30: If I do `New[AnyPrefix]()` this will return an id of the form `_123` right? That seems wrong (i.e. it would be missing the prefix completely)

typeid/typeid-go/typeid.go
```diff
@@ -62,96 +67,143 @@ func (tid TypeID) UUIDBytes() []byte {
...
-func From(prefix string, suffix string) (TypeID, error) {
-	if err := validatePrefix(prefix); err != nil {
-		return Nil, err
+func From(prefix string, suffix string) (TypeID[AnyPrefix], error) {
+	return from(AnyPrefix(prefix), suffix)
+}
+
+// FromPrefix returns a new TypeID with the given prefix and a random suffix.
+// If you want to create an id without a prefix, pass an empty string.
+func FromPrefix(prefix string) (TypeID[AnyPrefix], error) {
```
mikeland73, line 83: I might call this `NewWithPrefix` or `WithPrefix`

`from` implies you are creating the ID from the prefix which is not 100% true. (See how `From` and `FromString` are different)

typeid/typeid-go/typeid.go
mikeland73: Additionally, I just don't understand the value of this. You can still templatize the struct without turning the prefix into a generic type which makes the logic everywhere more complex.

mikeland73 (changes requested):
I agree with getting rid of untyped, but I'm not sure the approach in this PR is best. `AnyPrefix` feels pretty complex and IMO the language is a bit confusing. I don't understand why the prefix field in the struct is now a generic. What value do we get from that?

A few questions/comments:

* `New[AnyPrefix]` feels awkward. (In fact I don't think it should work at all. What's the point of using typeid with a missing prefix?) I would prefer to limit this functions to only typed typeids.
* I find all the prefix naming really confusing. There's a prefix field with a prefix pseudo-type (the generic P value in TypeID), also a prefix interface which has a function that is also called prefix(). 
* The current approach makes something like https://github.com/jetpack-io/opensource/pull/185 impossible to do. Because the constructor now requires a prefix type, instead of the ID type.

When you mentioned you were going to merge typed and untyped, I assumed you were going to get rid of untyped but keeping the existing structure until we decided what the best interface is. This feels pretty opinionated, which is OK, but it seems like a lot of code changes without fixing the main pain points we had discussed, specifically:

* The zero (nil) ID value should match the default value
* Reduce boilerplate when creating a new ID type
* Ideally having a single type (so the generic type is the same type that you use in your code)
* Reduce or eliminate untyped use of the library (avoid having a "wrong" way to use it). For example forcing people to declare a type in order to use.

loreto:
> I agree with getting rid of untyped, but I'm not sure the approach in this PR is best. `AnyPrefix` feels pretty complex and IMO the language is a bit confusing. I don't understand why the prefix field in the struct is now a generic. What value do we get from that?
> 
That doesn't matter. I can make the prefix field inside the struct a string and keep the generic on the struct itself. Everything else can continue to work as is (nothing really depends on the prefix field being of type P vs string)

> A few questions/comments:
> 
> * `New[AnyPrefix]` feels awkward. (In fact I don't think it should work at all. What's the point of using typeid with a missing prefix?) I would prefer to limit this functions to only typed typeids.
Without this we cannot parse typeids "dynamically". Suppose you have a CSV file with a bunch of typeids of different types. You know they are typeids, but apriori, you don't know the types. You need something like `AnyPrefix` to be able to do this.

> * I find all the prefix naming really confusing. There's a prefix field with a prefix pseudo-type (the generic P value in TypeID), also a prefix interface which has a function that is also called prefix().
Open to naming suggestions.

> * The current approach makes something like [[typeid-go][RFC] Use struct with tags #185](https://github.com/jetpack-io/opensource/pull/185) impossible to do. Because the constructor now requires a prefix type, instead of the ID type.
Once I change the prefix field back to a string, your approach is still possible. All this is enforcing is that the generic type implements the `Prefix` interface. AKA, if we rename `Type()` in TypeID to `Prefix()`, all typeids meet the requirements of the interface (we would have to have an implementation where `Prefix()` reads from the struct tag, but I'm assuming that's possible)

> 
> When you mentioned you were going to merge typed and untyped, I assumed you were going to get rid of untyped but keeping the existing structure until we decided what the best interface is. This feels pretty opinionated, which is OK, but it seems like a lot of code changes without fixing the main pain points we had discussed, specifically:
> 
> * The zero (nil) ID value should match the default value
This is coming in a follow up PR (I mentioned it in the PR description, it was already getting too big)
> * Reduce boilerplate when creating a new ID type
> * Ideally having a single type (so the generic type is the same type that you use in your code)
This is true with this PR. In fact, it's the main problem it solves. Why do you see it as not being the case?
> * Reduce or eliminate untyped use of the library (avoid having a "wrong" way to use it). For example forcing people to declare a type in order to use.
Also true. There's only the generic version now. What am I missing?

typeid/typeid-go/typeid.go
loreto: No, it would return `123` with no underscore in front (unless you're saying I have a bug in handling the empty prefix?). We do allow for an empty prefix in the spec and in all other implementations in different languages.

typeid/typeid-go/typeid.go
loreto: Mentioned it in the other response, but I'll change the prefix by to a `string`, and I think that's the only issue. I believe `TypeID` itself can be compatible with the `Prefix` interface, and by extension, compatible with your PR. Tell me if I'm misunderstanding something though.

loreto:
Thinking about this more: while I can turn the prefix field back to a string, simplify some naming, and make sure the interface is compatible with your PR ... I think a lot the complexity is coming from the use of Generics. 

So now I'm leaning a completely opposite way: let's get rid of the generics and have a simple, easy to understand TypeID struct. For compile time safety people can use type definitions or we can make it easy to create subtypes via struct embeddings (but without generics).

mikeland73:
> Thinking about this more: while I can turn the prefix field back to a string, simplify some naming, and make sure the interface is compatible with your PR ... I think a lot the complexity is coming from the use of Generics.

Totally agree! I was thinking about this, we don't need generics for the struct. I still think generics will benefit the interface in order to create a user defined type without an awkward interface.

> So now I'm leaning a completely opposite way: let's get rid of the generics and have a simple, easy to understand TypeID struct. For compile time safety people can use type definitions or we can make it easy to create subtypes via struct embeddings (but without generics).

Agree with this. I did come up with one benefit to your approach of not storing the prefix string which is memory use. Specifically, storing the prefix over and over again is wasteful (though I assume golang optimizes this a bit maybe?) By getting the prefix from the type we would save on memory costs. One question here is if we can support the dynamic case you mentioned, specifically when a data source has IDs with many prefixes (or unknown prefix).

mikeland73:
I was messing around with the idea of not using generics, and on second thought not sure we can. Specifically `UnmarshalText` needs generics in order to validate the prefix (because embedded struct doesn't have access to parent struct). I'll share a PR with some explorations.

loreto:
Rephrasing slightly, I think we can remove generics for the type, and only use generic in a few limited places in the "constructors". I'll send you a version of what I'm thinking this weekend.

loreto:
@mikeland73 PTAL. I basically rewrote the entire PR. The TypeID struct no longer uses generics. The constructors do.

I think the interface is really nice once you define an embedded type. In particular take a look at the README and/or `subtype_example_test.go` to see how you use it.

mikeland73:
> Rephrasing slightly, I think we can remove generics for the type, and only use generic in a few limited places in the "constructors". I'll send you a version of what I'm thinking this weekend.

Yeah, we're on same page about generics, but I ran into an issue. See below:

What I ran into (at least with my implementation) is that `UnmarshalText` can't validate the type unless we parametrize the type (because `UnmarshalText` has `TypeID` receiver which has no access to the parent struct that embeds `TypeID`). 

If we decide to go the struct tag approach, this is solvable by using this slightly uglier interface:


```golang
type OrgID struct {
	typeid.TypeID[OrgID] `prefix:"org"`
}
```

By passing the struct itself as type T, then `UnmarshalText` can be rewritten as such

```golang
func (tid *TypeID[T]) UnmarshalText(text []byte) error {
  var newID T
  prefix := typeFromTag(newID)  // returns the type
  ...
}
```

 At this point, I still slightly prefer the tag approach but my convictions is weaker because the interface is starting to get pretty ugly. Either way, we may need to reintroduce generics into the struct because of this issue

I haven't reviewed the most recent changes. Does your `UnmarshalText` ensure the correct prefix for the type?

mikeland73:
Will review your changes in a bit

typeid/typeid-go/subtype.go
```diff
@@ -0,0 +1,49 @@
+package typeid
+
+// Subtype is an interface used to create a more specific subtype of TypeID
+// For example, if you want to create an `OrgID` type that only accepts
+// an `org_` prefix.
+type Subtype interface {
+	AllowedPrefix() string
+	isTypeID() bool // Private to only allow implementation by embedding TypeID
```
mikeland73, line 8: This doesn't actually prevent outside implementations. But should we care? If a library user wants to implement their own version, as long as it has the `init` function it should work right?

typeid/typeid-go/subtype.go
```diff
@@ -0,0 +1,49 @@
...
+	// "mutation" should return a copy with the modifications instead of modifying
+	// the original. We make an exception for this *private* method, because
+	// sometimes we need to modify the fields in the process of initializing
+	// a new subtype.
+	tid.prefix = prefix
+	tid.suffix = suffix
+}
+
+func (tid TypeID) isTypeID() bool {
+	return true
```
mikeland73, line 29: If you really wanted to limit concrete implementations of this function return a private type here.

mikeland73 (approved):
This seems great to me, but take a look at `UnmarshalText` comments

gcurtis (commented):
Just so we don't keep going back and forth forever, I’ll summarize my thoughts on the various approaches and then leave it to you.

- I liked the original implementation of this PR the most. Declaring new ID types only took 2 lines of code - one for the type declaration and one for the `Prefix` method implementation. The Encore blog post also suggests that this approach works well in practice. There are two ways we could go about the final implementation, and I think either would work well:
    - Unify untyped and typed IDs. Pro: this is a simpler API. Con: typed prefixes need to be repeated for every ID.
    - Keep untyped and typed separate (but perhaps in the same package). Pro: prefixes are stored out-of-band via types, which can give non-negligible memory savings for lots of IDs. Con: larger API.
- I’m not a fan of the struct tag approach. It doesn’t feel like Go and requires roughly the same amount of typing. It requires reflection which means constructing typeids will always require a heap allocation.
- The current implementation of this PR feels similar to the previous iteration, but adds a bunch of complexity that IMO doesn’t buy us much. The function signature of `New` in particular is unclear (I had to read through the source to know what PT was). Looking at benchmarks, heap allocations  went up and parsing/encoding was 20-50% slower compared to the previous iteration.

The last thing I'll mention is plain UUIDs are nice because they can be stack-allocated and are a fixed size, so they play well with caching/locality. If we want to make TypeIDs a UUID alternative, we should be aware of any API choices that will block our ability to make optimizations. For example, we probably want to avoid using interfaces in a way that will require IDs to always be moved to the heap.

typeid/typeid-go/subtype.go
loreto: Why doesn't it prevent outside implementations? I thought it did.

See:
image

loreto:
> Just so we don't keep going back and forth forever, I’ll summarize my thoughts on the various approaches and then leave it to you.
> 
> * I liked the original implementation of this PR the most. Declaring new ID types only took 2 lines of code - one for the type declaration and one for the `Prefix` method implementation. The Encore blog post also suggests that this approach works well in practice. There are two ways we could go about the final implementation, and I think either would work well:
>   
>   * Unify untyped and typed IDs. Pro: this is a simpler API. Con: typed prefixes need to be repeated for every ID.
>   * Keep untyped and typed separate (but perhaps in the same package). Pro: prefixes are stored out-of-band via types, which can give non-negligible memory savings for lots of IDs. Con: larger API.
> * I’m not a fan of the struct tag approach. It doesn’t feel like Go and requires roughly the same amount of typing. It requires reflection which means constructing typeids will always require a heap allocation.
> * The current implementation of this PR feels similar to the previous iteration, but adds a bunch of complexity that IMO doesn’t buy us much. The function signature of `New` in particular is unclear (I had to read through the source to know what PT was). Looking at benchmarks, heap allocations  went up and parsing/encoding was 20-50% slower compared to the previous iteration.
> 
> The last thing I'll mention is plain UUIDs are nice because they can be stack-allocated and are a fixed size, so they play well with caching/locality. If we want to make TypeIDs a UUID alternative, we should be aware of any API choices that will block our ability to make optimizations. For example, we probably want to avoid using interfaces in a way that will require IDs to always be moved to the heap.

I'm hoping we can do an optimization pass later (for example, we could choose to implement the fields of TypeID all in a single fixed-size byte array).

Tell me more about this though:
> For example, we probably want to avoid using interfaces in a way that will require IDs to always be moved to the heap.

What part of the interface are you thinking forces usage of the heap?

pkg/sandbox/typeids/typeids.go
```diff
@@ -1,26 +1,19 @@
...
-type organizationPrefix struct{}
-
-func (organizationPrefix) Type() string { return "org" }
+type ProjectID struct {
+	typeid.TypeID
+}
 
-type OrganizationID struct {
-	typeid.TypeID[organizationPrefix]
+func (ProjectID) AllowedPrefix() string {
```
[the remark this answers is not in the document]
loreto: For the base case (`typeid.TypeID`), all prefixes are allowed  (so `AllowedPrefix() != Prefix()`). That said, I can also check for that via a type switch. I'd be happy with that approach too. Do you prefer to do the type switch so we can make this `Prefix()`?

pkg/jetcloud/client.go
```diff
@@ -44,14 +44,15 @@ func (c *client) endpoint(path string) string {
 }
 
 func (c *client) newProjectID(ctx context.Context, tok *session.Token, repo, subdir string) (typeids.ProjectID, error) {
+	var nilID typeids.ProjectID
```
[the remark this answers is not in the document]
loreto: Yes. Although, I don't think I can make the 'nil' ID be the same as the empty id after all. It works for the base `typeid.TypeID`, but it doesn't work for subtypes (where even the `nil` id of that type should have the type's prefix).

So I'm gonna reintroduce a `Nil[T]` constructor, and we can choose to use `*typeid.TypeID` in places in our code where we want to make it easy to return `nil`

typeid/typeid-go/constructors.go
```diff
@@ -0,0 +1,115 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/gofrs/uuid/v5"
+	"go.jetpack.io/typeid/base32"
+)
+
+// New returns a new TypeID of the given type with a random suffix.
+func New[T Subtype, PT subtypePtr[T]]() (T, error) {
```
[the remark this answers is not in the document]
loreto: If we keep this interface, I'll go ahead and add examples to the docs of every constructor.

pkg/jetcloud/client.go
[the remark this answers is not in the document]
gcurtis: I'm confused as to why it won't work for subtypes. Why would the nil ID need a prefix? Wouldn't the type ensure only the same subtypes can be compared?

In general I'm not a huge fan of nil IDs. If our IDs were integers or strings, for example, I don't think we'd be returning nil.

pkg/sandbox/typeids/typeids.go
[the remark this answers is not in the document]
gcurtis: I lean toward doing the type assertion, but not a huge deal.

gcurtis:
> What part of the interface are you thinking forces usage of the heap?

I don't know if anything in this PR's API does (I haven't really checked). I was just speaking more generally. As an example, if you have a function that takes an interface argument, that argument needs to be on the heap due to dynamic dispatch. There's no way to get around that unless you change the API again in a backwards-incompatible way.

You can build with the `-gcflags='-m=2'` flag and the compiler will print out which objects are heap-allocated and why. Some cases can be optimized later and some we wouldn't care about (allocating for an error doesn't matter). But it can at least give an idea.

typeid/typeid-go/constructors.go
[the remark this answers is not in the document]
mikeland73: I think the code is complicated but the examples make it clear. This is just an unfortunate syntax for a simpler concept (parametrizing a function in a way the pointer to the type implements an interface)

pkg/jetcloud/client.go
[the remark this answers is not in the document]
mikeland73: I agree with not using nils and just using zero types instead. 

Is this not just doing? 

```golang
var newID T
return newID
```

Or am I missing something?

mikeland73:
> I’m not a fan of the struct tag approach. It doesn’t feel like Go and requires roughly the same amount of typing. It requires reflection which means constructing typeids will always require a heap allocation.

I feel less strongly about the tag approach, but I do feel that `New[T]` should return a `T`. Having 2 types I think is a suboptimal interface. And using the prefix type to create the ID type adds an extra level of cognitive load. 

> heap allocation
Can you clarify what requires a heap allocation? Does doing `Var newID T` cause an allocation?

> The current implementation of this PR feels similar to the previous iteration, but adds a bunch of complexity that IMO doesn’t buy us much. The function signature of New in particular is unclear (I had to read through the source to know what PT was). Looking at benchmarks, heap allocations went up and parsing/encoding was 20-50% slower compared to the previous iteration.

`New` generics was hard to write, but it's easy to use and leads to nicer code.

I agree with @loreto , I would optimize for ergonomics over performance and we can optimize in a follow up. We are the primary customers of this API right now and I think in our projects ergonomics are more important. We can always optimize this later or if not possible, create an alternative implementation that performs better.

PR #192, opened 2023-11-15
Flatten /pkg/sandbox to just /pkg

loreto:
## Summary
Move all packages inside of `/pkg/sandbox/` to just `/pkg/`. The idea is to simplify things:
+ Start by creating packages in `/pkg/`
+ If we decide to publish them on their own, move them to the top level.

In the process I renamed the `typeids` package to just `id`. I was finding the collision with `typeid` too confusing, and I think being able to do `id.OrgID` in the code is nicer.

## How was it tested?
`devbox run test`


loreto:
@mikeland73 I think the build fails here because of the "double" publish problem: I first need to check in this PR, so that the individual repos will be published, and _then_ the go mod will work. Is there another way around that?

mikeland73 (approved):
We should probably stop enforcing go mod sync

gcurtis (approved):
I like the typeids → id rename.

PR #199, opened 2023-11-21
[easy][auth] Add redirect_uri to auth

mikeland73:
## Summary

I think this was previously working because ory automatically redirects to the whitelisted redirect URI if there's only one. But once we added a second one it no longer redirects automatically and requires the param.


loreto (approved):
LGTM

PR #200, opened 2023-11-22
[opensource] Improve root-level devbox.json

loreto:
## Summary
Improves the scripts in our root-level devbox.json:
- Each script now prints what directory it's working on
- Linting by default tries to fix issues when it can
- Added a `fmt` script.

## How was it tested?
Ran locally

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -9,12 +9,13 @@
     ],
     "scripts": {
       "go-mod-sync": [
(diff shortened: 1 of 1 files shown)


mikeland73 (approved):
I'm always conflict between having a separate `lint` vs `fmt` especially when `lint` has `--fix` in it.

loreto:
Checked it in, but are you thinking that `lint` should also run `go fmt` and we get rid of a top-level `fmt` script?

PR #201, opened 2023-11-27
[api secrets] add api proto clients

savil:
## Summary

Introduces `pkg/api` so that our opensource projects like `envsec` can invoke
`api.jetpack.io` using the generated proto bindings in golang.

We could improve the `devbox run gen` script by making a bash-script that has a list of whitelisted directories, but for now just did the simplest possible change.

Question:
Do we want this to be a top-level `api` folder?

NOTE: the `go.work.sum` change is unrelated but its been showing up for me so committed it.

## How was it tested?

Ran `devbox run gen`


loreto (approved):
LGTM

PR #203, opened 2023-11-29
Improvements to typeid-sql

loreto:
## Summary
Improvements to typeid-sql:
- Provide example files for both the text version and the compound version of typeid
- Remove the UUIDv7 check, sometimes people want to store other UUIDs in typeid format
- Rename the equality operator from `=` to `===` (and `!==`). Unfortunately some libraries
  give special meaning to `=`, and that breaks functionality like `supabase db diff`

## How was it tested?
Ran supabase tests


LucilleH (approved):
Wow `===` haha nice!

PR #205, opened 2023-11-30
[envsec] use ENVSEC_JETPACK_API_ENDPOINT to override

savil:
## Summary

Also adds `devbox run login-dev` as a convenience command to get the dev 
access-token and id-token locally.

NOTE: 
here, the env-var's default value is the prod value. This is consistent with ENVSEC_CLIENT_ID and ENVSEC_ISSUER behavior (but different from what we do in apisvc)

## How was it tested?

did `devbox run login-dev` and verified the access token

### DIFF
--- a/envsec/envsec.go
+++ b/envsec/envsec.go
@@ -5,6 +5,7 @@ package envsec
 
 import (
 	"context"
+	"os"
 	"path"
 
 	"github.com/pkg/errors"
@@ -120,9 +121,6 @@ type JetpackAPIConfig struct {
 // prodJetpackAPIEndpoint is used for production.
 const prodJetpackAPIEndpoint = "https://api.jetpack.io"
 
-// localhostJetpackAPIEndpoint is used for local development.
-// const localhostJetpackAPIEndpoint = "http://localhost:8080"
-
 // JetpackAPIStore implements interface Config (compile-time check)
 var _ Config = (*JetpackAPIConfig)(nil)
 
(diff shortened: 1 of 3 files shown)


savil:
Current dependencies on/for this PR:
* main
  * **PR #205** <a href="https://app.graphite.dev/github/pr/jetpack-io/opensource/205?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/opensource/205?utm_source=stack-comment).

savil:
@ipince I agree with you in general. That's easy to achieve for services. For a binary that we ship, I'd need to locate where we build it and set the env-vars there, _and_ modify the existing ones. So, for the time being, I was staying consistent with the current convention used in the binary.

envsec/envsec.go
```diff
@@ -131,9 +129,9 @@ func (c *JetpackAPIConfig) IsEnvStoreConfig() bool {
 }
 
 func NewJetpackAPIConfig(token string) *JetpackAPIConfig {
-	return &JetpackAPIConfig{
-		endpoint: prodJetpackAPIEndpoint,
-		// endpoint: localhostJetpackAPIEndpoint,
-		token: token,
+	endpoint := prodJetpackAPIEndpoint
+	if e := os.Getenv("ENVSEC_JETPACK_API_ENDPOINT"); e != "" {
+		endpoint = e
```
mikeland73, line 134: Use `envvar.Get`

PR #213, opened 2023-12-05
[envsec CLI]  use dev-apisvc for all envsec operations in dev-mode, including CreateProject 

savil:
## Summary

This PR lets us use dev-apisvc for all envsec operations in dev-mode, including
CreateProject (which this PR also hooks up)

## How was it tested?

1. `envsec init` in a new project. Saw `.jetpack.io/dev.project.json` created
2. `envsec ls` and `envsec set` do work

### DIFF
--- a/envsec/pkg/envcli/flags.go
+++ b/envsec/pkg/envcli/flags.go
@@ -56,8 +56,8 @@ func (f *configFlags) validateProjectID(orgID id.OrgID) (string, error) {
 	if err != nil {
 		return "", errors.WithStack(err)
 	}
-	jc := jetcloud.JetCloud{APIHost: build.JetpackAPIHost(), IsDev: build.IsDev}
-	config, err := jc.ProjectConfig(wd)
+	c := jetcloud.Client{ApiHost: build.JetpackAPIHost(), IsDev: build.IsDev}
+	config, err := c.ProjectConfig(wd)
 	if errors.Is(err, os.ErrNotExist) {
 		return "", fmt.Errorf(
 			"project ID not specified. You must run `envsec init` or specify --project-id in this directory",
--- a/envsec/pkg/envcli/flags.go
+++ b/envsec/pkg/envcli/flags.go
@@ -56,7 +56,7 @@ func (f *configFlags) validateProjectID(orgID id.OrgID) (string, error) {
 	if err != nil {
 		return "", errors.WithStack(err)
 	}
-	c := jetcloud.Client{ApiHost: build.JetpackAPIHost(), IsDev: build.IsDev}
+	c := jetcloud.Client{APIHost: build.JetpackAPIHost(), IsDev: build.IsDev}
 	config, err := c.ProjectConfig(wd)
 	if errors.Is(err, os.ErrNotExist) {
 		return "", fmt.Errorf(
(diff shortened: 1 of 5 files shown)


savil:
Current dependencies on/for this PR:
* main
  * **PR #213** <a href="https://app.graphite.dev/github/pr/jetpack-io/opensource/213?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/opensource/213?utm_source=stack-comment).

pkg/jetcloud/project.go
```diff
@@ -16,17 +16,25 @@ var ErrProjectAlreadyInitialized = errors.New("project already initialized")
...
 
 type projectConfig struct {
 	ProjectID id.ProjectID `json:"project_id"`
 	OrgID     id.OrgID     `json:"org_id"`
 }
 
-func InitProject(ctx context.Context, tok *session.Token, dir string) (id.ProjectID, error) {
+// JetCloud keeps configuration for the JetCloud API.
+// Could use a less stuttery name.
+type JetCloud struct {
```
mikeland73, line 28: Nit, call this struct API?

And field can be Host

envsec/pkg/envcli/flags.go
```diff
@@ -105,14 +107,25 @@ func (f *configFlags) genConfig(cmd *cobra.Command) (*CmdConfig, error) {
 		}
 	}
 
-	ssmConfig, err := awsfed.GenSSMConfigFromToken(ctx, tok, true /*useCache*/)
-	if err != nil {
-		return nil, errors.WithStack(err)
+	var store envsec.Store
+	if !build.IsDev {
```
mikeland73, line 111: This is ok in a pinch, but we probably want to add a env var that determines which backend is used

pkg/jetcloud/client.go
```diff
@@ -26,12 +25,13 @@ type errorResponse struct {
 	} `json:"error"`
 }
 
-func newClient() *client {
+func (jc *JetCloud) newClient() *client {
```
mikeland73, line 28: This is a bit fragile. Maybe we can merge the api object and the client into one?

mikeland73 (approved):
Comments are not blocking, we can clean up once we remove legacy code

envsec/pkg/envcli/flags.go
savil: Yeah, I think we need some design thought into how to specify the backend for the envsec project. 

It may be part of `envsec init`, and then stored in .... an `envsec.json`? 

I may be missing something but I don't see how env-var would be ideal, because the users of a particular envsec-project should use the same backend store (i.e. all members of a team should use the same backend store for their project).

pkg/jetcloud/project.go
savil: Yeah, I considered `API` as well, but it feels odd for `API` to then be writing local files like `project.json`.

What about `jetcloud.Client`, and then we merge the other `client` from above?

pkg/jetcloud/client.go
savil: yeah, lets merge now that we have another struct

PR #219, opened 2023-12-08
[envsec] Add CICD tests

mikeland73:
## Summary

This is a bit duplicated with monorepo tests, but because sometimes we forget to update dependencies it's still possible for the envsec build to break. This will let us know. 

## How was it tested?

Will test in CICD


loreto (approved):
LGTM

PR #220, opened 2023-12-08
[envsec] Fix envsec CICD

mikeland73:
## Summary

* lint is breaking because of missing file, but we don't really need to lint.


loreto (approved):
LGTM

PR #221, opened 2023-12-12
[RFC] Enpasulate envsec init into lib

mikeland73:
## Summary

Quick and dirty turning envsec into library. Only did `init` for now but if we agree on structure I'll move the rest.

Ideally everything except flags is in there. 

* Creates new `envsec` package
*  Adds struct with `NewProject` function
* Adds `--force` flag to init.

Thoughts or feedback?

## How was it tested?

`devbox init --force`


loreto (approved):
LGTM

PR #223, opened 2023-12-13
[RFC] Remove prefix param when possible

mikeland73:
## Summary

This change removes the prefix param from `FromUUID` and `FromUUIDBytes`. It follows the philosophy that the correct path should be easy (and the bad path should be less easy).

Previously to convert a UUID you must do:

```golang
// Previously (ugly)
typeid.FromUUID[idType](idType{}.Prefix(), uuid)

// Or potential bug, pass a string
typeid.FromUUID[idType]("typePrefix", uuid)

// After this change
typeid.FromUUID[idType](uuid)
```

Benefits:

* less code in the correct path.
* Much better type checking and less opportunity for miss-use. No way to pass a bad prefix.

Downside:

* AnyIDs are a bit longer and uglier.
* Extra functions

Open to renaming `AnyPrefixFromUUID` to something better. Maybe `FromUUIDWithPrefix` ? or `FromPrefixAndUUID` ?

## How was it tested?

* builds, tests

### DIFF
--- a/typeid/typeid-go/constructors.go
+++ b/typeid/typeid-go/constructors.go
@@ -9,6 +9,8 @@ import (
 	"go.jetpack.io/typeid/base32"
 )
 
+var ErrConstructor = errors.New("constructor error")
+
 // New returns a new TypeID of the given type with a random suffix.
 //
 // Use the generic argument to pass in your typeid Subtype:
@@ -100,11 +102,25 @@ func split(id string) (string, string, error) {
 
 // FromUUID encodes the given UUID (in hex string form) as a TypeID
 func FromUUID[T Subtype, PT SubtypePtr[T]](uidStr string) (T, error) {
+	if isAnyID[T]() {
+		var id T
+		return id, fmt.Errorf(
+			"%w: use FromUUIDWithPrefix(), FromUUID() is for Subtypes",
+			ErrConstructor,
+		)
+	}
 	return fromUUID[T, PT](defaultPrefix[T](), uidStr)
 }
 
 // FromUUIDBytes encodes the given UUID (in byte form) as a TypeID
 func FromUUIDBytes[T Subtype, PT SubtypePtr[T]](bytes []byte) (T, error) {
+	if isAnyID[T]() {
+		var id T
+		return id, fmt.Errorf(
+			"%w: use FromUUIDBytesWithPrefix(), FromUUIDBytes() is for Subtypes",
+			ErrConstructor,
+		)
+	}
 	uidStr := uuid.FromBytesOrNil(bytes).String()
 	return FromUUID[T, PT](uidStr)
 }
--- a/typeid/typeid-go/constructors.go
+++ b/typeid/typeid-go/constructors.go
@@ -98,8 +98,8 @@ func split(id string) (string, string, error) {
 	}
 }
 
-// FromUUID encodes the given UUID (in hex string form) as a TypeID with the given prefix.
-func FromUUID[T Subtype, PT SubtypePtr[T]](prefix string, uidStr string) (T, error) {
+// FromUUID encodes the given UUID (in hex string form) as a TypeID
+func FromUUID[T Subtype, PT SubtypePtr[T]](uidStr string) (T, error) {
 	uid, err := uuid.FromString(uidStr)
 	var nilID T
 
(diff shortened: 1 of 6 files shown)


typeid/typeid-go/prefix.go
```diff
@@ -22,6 +22,12 @@ type AnyID struct {
 	TypeID[AnyPrefix]
 }
 
+type DynamicPrefix string
```
loreto, line 25: Why do you need this? (how is it different from the Any Prefix?). I'm tempted to get rid of it unless I understand the use case.

typeid/typeid-go/constructors.go
```diff
@@ -98,22 +98,39 @@ func split(id string) (string, string, error) {
...
+func FromUUID[T Subtype, PT SubtypePtr[T]](uidStr string) (T, error) {
 	uid, err := uuid.FromString(uidStr)
 	var nilID T
 
 	if err != nil {
 		return nilID, err
 	}
 	suffix := base32.Encode(uid)
-	return from[T, PT](prefix, suffix)
+	return from[T, PT](nilID.Prefix(), suffix)
```
loreto, line 110: I had been using `defaultPrefix[T]()` to get the prefix of the passed in type:
```
        prefix := defaultPrefix[T]()
	suffix := base32.Encode(uid)
	return from[T, PT](prefix, suffix)
```

typeid/typeid-go/constructors.go
```diff
@@ -98,22 +98,39 @@ func split(id string) (string, string, error) {
...
+
+// FromUUIDBytes encodes the given UUID (in byte form) as a TypeID
+func FromUUIDBytes[T Subtype, PT SubtypePtr[T]](bytes []byte) (T, error) {
+	uidStr := uuid.FromBytesOrNil(bytes).String()
+	return FromUUID[T, PT](uidStr)
+}
+
+// AnyPrefixFromUUID encodes the given UUID (in hex string form) as a TypeID with the given prefix.
+func AnyPrefixFromUUID(prefix string, uidStr string) (AnyID, error) {
+	uid, err := uuid.FromString(uidStr)
```
loreto, line 121: nit, but if you keep around a private implementation the takes in both the generic and the prefix, you can use it as the single implementation for both the AnyPrefix case and the typed case:
```
func fromUUID[T Subtype, PT SubtypePtr[T]](prefix string, uidStr string) (T, error) {
  ... // implementation here
 }
 
 func FromUUID [T Subtype, PT SubtypePtr[T]](uidStr string) (T, error) {
   // throw error if dealing with Any, see what we do in New()
   prefix := defaultPrefix[T]()
   return fromUUID[T, PT](prefix, uidstr)
  }
  
  func FromUUIDWithPrefix(prefix string, uidStr string) (AnyID, error) {
    return fromUUID[AnyID](prefix, uidStr)
   }
```

Same for the UUIDBytes implementation

typeid/typeid-go/constructors.go
```diff
@@ -98,22 +98,39 @@ func split(id string) (string, string, error) {
...
+	if err != nil {
+		return AnyID{}, err
+	}
+	suffix := base32.Encode(uid)
+	return from[AnyID, *AnyID](prefix, suffix)
 }
 
 // FromUUID encodes the given UUID (in byte form) as a TypeID with the given prefix.
-func FromUUIDBytes[T Subtype, PT SubtypePtr[T]](prefix string, bytes []byte) (T, error) {
+func AnyPrefixFromUUIDBytes(prefix string, bytes []byte) (AnyID, error) {
```
loreto, line 131: Could we do: `FromUUIDBytesWithPrefix` and `FromUUIDWithPrefix` (matches the `WithPrefix` constructor we already have)

loreto (commented):
I'm cool with this change, see comments.

It is a breaking change so we would have to release v2 of typeid

typeid/typeid-go/prefix.go
mikeland73: oops, not used. I was just messing around with alternatives. Will rm

mikeland73:
> It is a breaking change so we would have to release v2 of typeid

We don't need to release right away right? We may want to put some more breaking changes before we bump the major version

mikeland73:
@loreto made the changes PTAL

typeid/typeid-go/constructors.go
```diff
@@ -98,8 +98,31 @@ func split(id string) (string, string, error) {
 	}
 }
 
-// FromUUID encodes the given UUID (in hex string form) as a TypeID with the given prefix.
-func FromUUID[T Subtype, PT SubtypePtr[T]](prefix string, uidStr string) (T, error) {
+// FromUUID encodes the given UUID (in hex string form) as a TypeID
+func FromUUID[T Subtype, PT SubtypePtr[T]](uidStr string) (T, error) {
+	return fromUUID[T, PT](defaultPrefix[T](), uidStr)
```
loreto, line 103: Could you add:
```
	if isAnyID[T]() {
		var id T
		return id, errors.New("constructor error: use FromUUIDWithPrefix(), FromUUID() is for Subtypes")
	}
```

Is just to avoid the case of someone calling `FromUUID[AnyID](...)`, that results in a typeid with no prefix, which is probably unintended.

Same in FromUUIDBytes()

loreto (approved):
LGTM

One last comment inline

PR #224, opened 2023-12-13
[typeid-go] Bump version to v2

mikeland73:


loreto (commented):
Does this work even if we don't yet officially publish a v2 release? (I'd love to hold off on publishing the release in case we want to make other breaking changes)

If it does, looks great!

mikeland73:
I was doing this in preparation for v2 tag, but I'll hold off.

In the meantime, I can update any users to newest  API by using `go get -u go.jetpack.io/typeid@3871c75`

PR #225, opened 2023-12-14
[envsec] better init and whoami

mikeland73:
## Summary

Better UX for `init` and `whoami`

* `init` now shows prompt `Initializing project for mike-test-team. Enter y/yes to continue`. Allows user to create or abort.
* `--force` skips prompt.
* whoami shows team name and email (added scopes)

## How was it tested?

```bash
envsec init
envsec init --force
envsec auth whoami
```


envsec/pkg/envcli/auth.go
```diff
@@ -80,39 +81,15 @@ func whoAmICmd() *cobra.Command {
...
-
-			return nil
+			return (&envsec.Envsec{
+				APIHost: build.JetpackAPIHost(),
+				Auth: envsec.AuthConfig{
+					ClientID: envvar.Get("ENVSEC_CLIENT_ID", build.ClientID()),
+					Issuer:   envvar.Get("ENVSEC_ISSUER", build.Issuer()),
+				},
+				IsDev:  build.IsDev,
+				Stderr: cmd.ErrOrStderr(),
```
savil, line 91: this looks copy-pasta from `envcli/init.go`. Can you pull into a helper function?

pkg/jetcloud/project.go
```diff
@@ -121,3 +131,22 @@ func (c *Client) configPath(wd string) string {
...
+		return err
+	}
+	fmt.Fprintf(
+		args.Stderr,
+		"Initializing project for %s. Enter y/yes to continue\n",
+		member.Organization.Name,
+	)
+	result := ""
+	fmt.Scanln(&result)
+	if result != "y" && result != "yes" {
```
savil, line 148: Can we also check `Y` and `Yes` and `YES`?

I also feel that ENTER should be treated as a yes.

savil: These prompts are usually of the form:
`Initializing project for %s. Do you want to continue? [Y/n]\n`
https://askubuntu.com/a/322601

which signifies that hitting ENTER will be treated as a yes, alongside uppercase and lowercase `y`

savil (approved):
nice!

envsec/devbox.json
```diff
@@ -22,6 +23,12 @@
         "devbox run build",
         "dist/envsec auth login",
       ],
+      "fmt": [
+        "find . -name '*.go' -not -wholename './gen/*' -exec gofumpt -w -l {} \\+",
+        "if [ -n \"$CI\" ]; then",
+        "  git diff --exit-code",
+        "fi"
+      ],
```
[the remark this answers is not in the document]
mikeland73: This is not called anywhere (but it's there if we do want to enforce it)

PR #226, opened 2023-12-14
[envsec] Don't require git repo

mikeland73:
## Summary

TSIA

In a follow up, I'll improve project name when not in github repo.

## How was it tested?

`envsec init` (not in repo)

### DIFF
--- a/pkg/jetcloud/git.go
+++ b/pkg/jetcloud/git.go
@@ -20,7 +20,7 @@ func gitRepoURL(wd string) (string, error) {
 	if err != nil {
 		// Only check this if first call errors to avoid the extra call.
 		if !isInGitRepo(wd) {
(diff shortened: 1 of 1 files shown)


loreto (approved):
LGTM ... although once this is in, I have two questions:
1. How will GetOrCreate work? I thought the name of the repo is a key parameter to it
2. Will we add the "link" functionality? Or rather: in a different computer how do I associate a directory with an existing project that has no git repo?

mikeland73:
@loreto I was thinking of a prompt that allows user to pick an existing project or create a new one. Something like 

```
envsec init
* project 1
* project 2
* project 3
* Create new project
```

Something along those lines. 

So we may not end up needing `GetOrCreate` TBD

PR #232, opened 2023-12-19
[envsec] Implement init survey

mikeland73:
## Summary

This implements a simplified version of the init survey. Basically:

```
// flow:
// 0. Ask if you want to overwrite existing config [y/N]
// 1. Link to an existing project? [Y/n]
// 2a. What project would you like to link to? (sorted by repo/dir match)
// 2b. What’s the name of your new project?
```
 
A few other cleanup:

* Got rid of `jetcloud` package. Split into `api` and `git`
* Use transport to add auth token. (@savil we should probably do the same for secrets API).

## How was it tested?

`envsec init`


envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
+package flow
```
loreto, line 1: initflow?

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/samber/lo"
+	"go.jetpack.io/envsec/internal/git"
+	"go.jetpack.io/pkg/api/api"
+	projectsv1alpha1 "go.jetpack.io/pkg/api/gen/priv/projects/v1alpha1"
+	"go.jetpack.io/pkg/auth/session"
+	"go.jetpack.io/pkg/id"
+	"go.jetpack.io/typeid"
+)
+
+// flow:
```
loreto, line 22: I'm gonna leave a bunch of comments w/ suggestions for the flow, but feel free to postpone some of those to other PRs (if you agree with them). I think this is a great improvement already so let's check it in quickly and iterate on it.

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return id.ProjectID{}, err
+	}
+	if linkToExisting {
+		return i.showExistingListPrompt(ctx)
+	}
+	return i.createNewPrompt(ctx)
+}
+
+func (i *Init) overwriteConfigPrompt() (bool, error) {
```
loreto, line 60: Instead of overwrite, I think you can make this a confirmation prompt for all projects.

If project has never been initialized the prompt is:
`Set up project in "<dir>"? [Y/n]` defaulted to yes.

If the project already exists, then you do:
`Project already exists. Reset project in "<dir>"? [y/N]` defaulted to no

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+		if err != nil {
+			return id.ProjectID{}, err
+		}
+		if !overwrite {
+			return id.ProjectID{}, errors.New("aborted")
+		}
+	}
+
+	// TODO: printTeamNotice will be a team picker once that is implemented.
+	if err := i.printTeamNotice(ctx); err != nil {
```
loreto, line 47: Team -> Org

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (i *Init) printTeamNotice(ctx context.Context) error {
+	member, err := i.Client.GetMember(ctx, i.Token.IDClaims().Subject)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(
+		os.Stderr,
+		"Initializing project for %s\n",
```
loreto, line 74: `Initializing project in org "<org>"\n`

loreto: I'm unsure if we should print just `<org name>` or `<org name> (<org id>)`

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+		return id.ProjectID{}, err
+	}
+
+	directory, err := git.GitSubdirectory(i.WorkingDir)
+	if err != nil {
+		return id.ProjectID{}, err
+	}
+
+	sort.SliceStable(projects, func(i, j int) bool {
+		return projects[i].GetRepo() == repo &&
```
loreto, line 108: Nit. 

In the sort order it'd be nice to have:
1. Matching repo and subdir sorted by reverse chronological order (I think that's what you already have, server should be returning in reverse chronological)
2. Matching repo but different subdir by reverse chronological order (missing)
3. Everything else

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+			projects[i].GetDirectory() == directory
+	})
+
+	prompt := promptui.Select{
+		Label: "What project would you like to link to",
+		Items: lo.Map(projects, func(p *projectsv1alpha1.Project, _ int) string {
+			item := strings.TrimSpace(p.GetName())
+			if item == "" {
+				item = "unnamed project"
```
loreto, line 117: `untitled`?

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+			}
+			if p.GetRepo() != "" {
+				item += " repo: " + p.GetRepo()
+			}
+			if p.GetDirectory() != "" && p.GetDirectory() != "." {
+				item += " dir: " + p.GetDirectory()
+			}
+			return item
```
loreto, line 125: Wondering if we should also be displaying the project id. Does it get too big?

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (i *Init) createNewPrompt(ctx context.Context) (id.ProjectID, error) {
+	prompt := promptui.Prompt{
+		Label:   "What’s the name of your new project",
+		Default: filepath.Base(i.WorkingDir),
+		Validate: func(name string) error {
+			if name == "" {
+				return errors.New("project name cannot be empty")
+			}
```
loreto, line 144: Nit we might have a few more restrictions on the name that might be good to enforce client-side (and document in the endpoint if we haven't). I.e. same rules as k8s names? Size limit? Etc

pkg/api/api/projects.go
```diff
@@ -0,0 +1,54 @@
+package api
```
loreto, line 1: The double `api/api` directory is weird to me.

Couldn't we do:

```
pkg/api/
  projects.go
  gen/...
```

I'm also tempted to call it `apiclient` instead of just `api`? 🤷

pkg/api/api/projects.go
```diff
@@ -0,0 +1,54 @@
...
+	projectsv1alpha1 "go.jetpack.io/pkg/api/gen/priv/projects/v1alpha1"
+	"go.jetpack.io/pkg/api/gen/priv/projects/v1alpha1/projectsv1alpha1connect"
+	"go.jetpack.io/pkg/id"
+)
+
+func (c *Client) ListProjects(
+	ctx context.Context,
+	orgID id.OrgID,
+) ([]*projectsv1alpha1.Project, error) {
+	memberResponse, err := projectsv1alpha1connect.NewProjectsServiceClient(
```
loreto, line 16: Do you really want to create a new service client in each request?

Why not initialize these as fields of the `Client` struct on `NewClient`

pkg/api/api/client.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func NewClient(host string, token *session.Token) *Client {
+	return &Client{
+		Host:       host,
+		httpClient: &http.Client{Transport: &transport{token}},
+	}
+}
+
+type transport struct{ token *session.Token }
```
loreto, line 23: I don't think we should define our own transport. Why doesn't `oauth2.NewClient()` or `oauth2.Config.Client()` (for the first one, we could just pass a static token source if all you want to do is pass the token we already have, eventually, if we setup the config and create the client from there, it would handle a token refresh automatically, but we don't need to do that yet)

envsec/internal/flow/init.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	linkToExisting, err := i.linkToExistingPrompt()
+	if err != nil {
+		return id.ProjectID{}, err
+	}
+	if linkToExisting {
+		return i.showExistingListPrompt(ctx)
+	}
+	return i.createNewPrompt(ctx)
+}
```
loreto, line 58: Where do we print the final confirmation that the project was in fact linked `Linked to project <...>` or created `Created project <...> in org <...>`

envsec/internal/flow/init.go
mikeland73: I was thinking we can use this for more than just init?

envsec/internal/flow/init.go
mikeland73: Sounds good. Yeah, I'm thinking we should iterate instead of getting it perfect. I used it a bunch and it feels pretty good (maybe 80% of where it needs to be)

envsec/internal/flow/init.go
mikeland73: I thought we use the word team?

envsec/internal/flow/init.go
mikeland73: It gets somewhat hard to read, but I can add it.

envsec/internal/flow/init.go
mikeland73: follow up?

pkg/api/api/projects.go
mikeland73: Sure. I originally called it `client` but I really dislike the name.

I think putting it in top level sounds good.

pkg/api/api/projects.go
mikeland73: Sure, don't feel strongly.

envsec/internal/flow/init.go
loreto: In the UI we do use the word team currently ... I've been thinking we should go ahead and change it to Organization in the UI too. I'm getting confused between the use of Team some times and the use of Organization other times.

loreto (approved):
LGTM to unblock, since a lot of the comments could be follow ups for a new PR (and it'd be good to do a new bug bash w/ this flow sooner rather than later)

PR #237, opened 2023-12-20
[envsec] fix dev

mikeland73:
## Summary

We're accidentally checking wrong config.

### DIFF
--- a/envsec/pkg/envcli/flags.go
+++ b/envsec/pkg/envcli/flags.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/envsec"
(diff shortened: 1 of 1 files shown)


savil (approved):

savil:
1. I'm pretty worried this most basic functionality was broken. We should do sanity tests of dev mode before landing any PRs
2. Consider making the fields in Envsec private and adding a constructor function
to ensure we set some required fields.

mikeland73:
> I'm pretty worried this most basic functionality was broken. We should do sanity tests of dev mode before landing any PRs

Yeah. I was mostly testing `init` and forgot about this change. It was part of a refactor and my bad. Agree that having some test scripts would be food.

> Consider making the fields in Envsec private and adding a constructor function
to ensure we set some required fields.

Agreed.

PR #240, opened 2023-12-20
[auth] Add multi-session support for auth pkg

mikeland73:
## Summary

Adds multi-session support to auth. 

* Every time a user logs in, we prepend the token to the token slice. We always dedup by subject, so if there's any existing token that matches that subject, we remove it.
* Added `GetSessions` function. `GetSessions` returns refreshable tokens, instead of tokens. This allows the caller to inspect each token to find which one they want without having to refresh all. It also allows them to do useful stuff if only some tokens refresh successfully. 
* Session tokens are saved in a json file that contains a version.
* Logout removes all tokens.
* These changes are fully backwards compatible (which the exception that they will consider current users to be logged out, so they will require a single login on upgrade)

This will support allowing envsec to:

* Initialize a project in any org you are logged in to.
* Automatically use the correct token if you are logged in to multiple orgs and have projects in multiple orgs.

## How was it tested?

`envsec auth login`

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -2,7 +2,6 @@ package auth
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


pkg/auth/internal/tokenstore/tokenstore.go
```diff
@@ -7,10 +7,17 @@ import (
...
 
+const storeDataVersion = "1"
+
 type Store struct {
 	rootDir string
 }
 
+type storeData struct {
+	Version string           `json:"version"`
+	Tokens  []*session.Token `json:"tokens"`
```
savil, line 18: Would be nice to comment here that the order matters: tokens in front are given priority, and a new token is pre-pended to the existing list

pkg/auth/auth.go
```diff
@@ -97,6 +97,51 @@ func (c *Client) GetSession(ctx context.Context) (*session.Token, error) {
 	return tok, nil
 }
 
+// GetSessions returns all session tokens. It attempts to refresh any
+// expired tokens, and returns an error if any tokens are unable to be
+// refreshed. If there are errors, it still returns all tokens, some of which
+// may be expired.
+func (c *Client) GetSessions(ctx context.Context) ([]*session.Token, error) {
```
savil, line 104: where is this used?

pkg/auth/internal/tokenstore/tokenstore.go
```diff
@@ -53,3 +97,25 @@ func (s *Store) DeleteToken(issuer string, clientID string) error {
...
+}
+
+func (sd *storeData) addToken(tok *session.Token) {
+	tokens := []*session.Token{tok}
+	for _, t := range sd.Tokens {
+		if t.IDClaims().Subject != tok.IDClaims().Subject {
+			tokens = append(tokens, t)
+		}
+	}
+	sd.Tokens = tokens
```
savil, line 120: The order seems to matter and have product implications in terms of what is returned from ReadToken. Given that:

Can we distinguish login-writes from refresh-writes? 

I can understand wanting to move login-write-tokens to the front for priority. However, its unclear to me that refresh-write-tokens should move to the front, or whether they should be kept in their current position

savil (commented):
This looks great - one question about whether we want to _always_ prepend to the start of the tokens list for every write, or whether it makes sense to distinguish login from refresh writes?

savil (approved):
spoke offline - will keep refresh writes in-place, approving pre-emptively

pkg/auth/auth.go
```diff
@@ -97,6 +97,51 @@ func (c *Client) GetSession(ctx context.Context) (*session.Token, error) {
...
+
+	var refreshErrors []error
+	for idx, tok := range tokens {
+		if !tok.Valid() {
+			tokens[idx], err = c.refresh(ctx, tok)
+			refreshErrors = append(refreshErrors, err)
+		}
+	}
+
+	return tokens, errors.Join(refreshErrors...)
```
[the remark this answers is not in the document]
mikeland73: The idea behind this API is to allow the caller to still display useful into even if error. For example, you have 3 tokens, 2 work and one failed to refresh. The caller wants to show what orgs are available. In that case, it can display:

* org 1
* org 2
* org 3 (requires re-login)

instead of just erroring out and requiring full login no matter what.

returning a slice of tokens and a slice or errors would be very un-go'ish. Another alternative is to return something like:

```golang
type TokenSource interface {
	Token() (*Token, error)
}
```

(this interface is actually part of oauth2 package)

pkg/auth/auth.go
mikeland73: not currently used, but the plan is for envsec to use it to show what orgs you can use to initialize a project.

pkg/auth/internal/tokenstore/tokenstore.go
mikeland73: will do, this is a bug in my implementation. Refreshing a token should not change the order, only logins should.

PR #247, opened 2024-01-03
[envsec] Add info command and auto-switch orgs

mikeland73:
## Summary

This adds `envsec info` command that shows current project. It automatically detects the correct org.
We also now pick the correct credentials for the current project. This allows multiple credentials to work without having to re-log in.

## How was it tested?

`envsec info`
image


loreto:
I know you asked for thoughts on this UX on slack .... I've been mulling on it. Assuming this is exposed to users, I think I'd prefer something like `envsec info` over `envsec projects current` (the last one feels awkward to me for some reason)

mikeland73:
@loreto changed to `info` PTAL

mikeland73:
I also added code to select the correct credentials for users that are logged in to multiple accounts.

loreto (approved):
LGTM

PR #248, opened 2024-01-04
[envsec] Use survey intead of promptui

mikeland73:
## Summary

Replace `promptui` with `survey`. Survey is archived but more recently used. It seems more stable. Other option is https://github.com/charmbracelet/huh but it is very large.

Removed repo/dir from prompt to make lines more readable. See https://jetpack-io.slack.com/archives/C05NSEPE21H/p1704327292057999

## How was it tested?

image


LucilleH (approved):
I like survey better 👍

PR #252, opened 2024-01-06
[envsec] Add List function to envsec library

mikeland73:
## Summary

This adds a `List` and `PrintEnvVars`  functions to the envsec library. It allows a 3rd party (e.g. devbox) app to use envsec functionality more easily.

I refactored stores as much as I could to make this easy to use, but it's still a bit tangled. Specifically, the envsec package uses the jetpack API internally a bunch. In theory, we could make most of these functions part of the `store`. That way other stores can implement their own `init` or `whoami` specifics depending on their capabilities. This is left for later.

Once this is merged, devbox will be able to:

* Call `envsec.List` in order to fetch secrets (instead of the current implementation which calls the envsec CLI).
* Implement a `devbox secrets list` command that fetches and prints secrets.

## How was it tested?

```bash
envsec ls
envsec foo=123
envsec ls
```


envsec/pkg/stores/ssmstore/ssmstore.go
```diff
@@ -30,14 +32,18 @@ func newSSMStore(ctx context.Context, config *SSMConfig) (*SSMStore, error) {
 	return store, nil
 }
 
-func (s *SSMStore) List(ctx context.Context, envID EnvID) ([]EnvVar, error) {
+func (s *SSMStore) Identify(context.Context, *envsec.Envsec, *session.Token) {
+	// TODO: implement
```
savil, line 36: maybe add a `panic("TODO: implement SSMStore.Identify")` ?

savil (approved):
awesome!

envsec/pkg/stores/ssmstore/ssmstore.go
mikeland73: I don't really want it to panic since it does identity in a different way (via standard AWS credential chain). If we make it panic here it will break.

PR #255, opened 2024-01-08
[envsec] Implement Set,Delete as library functions

mikeland73:
## Summary

Implements:

* Set, SetMap, SetFromArgs library functions 
* Implements `DeleteAll`
* Clean up logic in `genConfig`
* Moved some output from stdout to stderr.

## How was it tested?
`ls`, `set`, `rm`, `upload`, `download`

### DIFF
--- a/envsec/pkg/envcli/flags.go
+++ b/envsec/pkg/envcli/flags.go
@@ -82,9 +82,9 @@ func (f *configFlags) validateProjectID(orgID id.OrgID) (string, error) {
 }
 
 type CmdConfig struct {
-	Store    envsec.Store
-	EnvID    envsec.EnvID
-	EnvNames []string
+	envsec   *envsec.Envsec
+	envID    envsec.EnvID
+	envNames []string
 }
 
 func (f *configFlags) genConfig(cmd *cobra.Command) (*CmdConfig, error) {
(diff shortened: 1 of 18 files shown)


mikeland73:
@savil @loreto I know this is a bit large, but it's a blocker for devbox work. It's fairly straight forward plus clean up!

envsec/pkg/envsec/set.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	insertedNames := lo.Keys(envMap)
+	return tux.WriteHeader(e.Stderr,
+		"[DONE] Set environment %s %v in environment: %s\n",
+		tux.Plural(insertedNames, "variable", "variables"),
+		strings.Join(tux.QuotedTerms(insertedNames), ", "),
+		strings.ToLower(envID.EnvName),
+	)
```
savil, line 34: I don't know if it makes sense to have the package write Info messages like this to stderr. 

The reason is that caller applications like Devbox may have their own libraries (like Devbox's `ux.Finfo`) for emitting information like this.

Don't want to block on this but could iterate on in the future

savil (approved):

envsec/pkg/envcli/flags.go
```diff
@@ -82,9 +82,9 @@ func (f *configFlags) validateProjectID(orgID id.OrgID) (string, error) {
 }
 
 type CmdConfig struct {
-	Store    envsec.Store
-	EnvID    envsec.EnvID
-	EnvNames []string
+	envsec   *envsec.Envsec
```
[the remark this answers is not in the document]
mikeland73: Returning the store felt like it was getting convoluted. It also required each caller to set the store individually. Long term this will need rethinking if we want to support more open source non-jetpack stores.

Longer term, I kinda imagine envsec working something like:

`envsec.New(ctx, store, envsec.WithAuthProvider(myAuthProvider))`

Which would allow you to mix and match stores and auth providers. Maybe stores have default auth providers as well. In that world, `genConfig` really just becomes a function to create a new envsec instance.

envsec/pkg/envsec/set.go
mikeland73: This is already doable as is, if devbox wants to emit its own messages it can simply pass dev null or a buffer as Stderr. This would silence all output. This way we don't need each user to the library to implement its own printing, they only do it if they want.

PR #257, opened 2024-01-10
[envsec] Implement download/upload as lib

mikeland73:
## Summary

* Implements `download` and `upload` as envsec lib functions.
* Adds format autodetect (if json) for both upload and download.
* Simplifies genConfig
* Simplified Envsec interface by adding EnvID to struct.
* Simplifies `List` by only listing single environment. We can decide if we want to support multi-environment later.

## How was it tested?

```
envsec download foo.json
envsec download foo.env
# modified both
envsec upload foo.json foo.env
```


envsec/pkg/stores/ssmstore/ssmstore.go
```diff
@@ -20,20 +23,25 @@ type SSMStore struct {
 // SSMStore implements interface Store (compile-time check)
 var _ envsec.Store = (*SSMStore)(nil)
 
-func New(ctx context.Context, config *SSMConfig) (*SSMStore, error) {
-	paramStore, err := newParameterStore(ctx, config)
+func (s *SSMStore) Identify(ctx context.Context, e *envsec.Envsec) (*session.Token, error) {
```
savil, line 26: call it `Init` or `New` since its identifying and setting the `store` field?

savil: that would work for JetStore as well

savil (approved):

envsec/pkg/stores/ssmstore/ssmstore.go
mikeland73: thoughts on `InitForUser`?  This is a bit of a weird function, it initializes but also returns the user.

savil: oh right, sgtm

PR #260, opened 2024-01-19
Sort secrets before displaying them

ipince:
## Summary
Otherwise the secrets are displayed in a non-deterministic order. It feels weird that repeated calls to `envsec ls` result in different outputs. Also alphabetizing makes it easier to find secrets in large lists (like dashboard).

## How was it tested?
In progress.

### DIFF
--- a/envsec/pkg/envsec/list.go
+++ b/envsec/pkg/envsec/list.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"slices"
 	"strings"
 
 	"github.com/olekukonko/tablewriter"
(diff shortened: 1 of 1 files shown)


mikeland73:
non-deterministic is not great, but not sure abc is best. Does gcp have a timestamp we could use instead?

Either way, this is probably best done server side and the client should respect the order the server returns.

ipince:
I agree we should include timestamps. I was thinking of adding that afterwards, and displaying on a third column. That said, I think abc order is better, while still displaying last updated time. (and maybe later accept a `--sort` flag maybe, but that seems like too much for now)

ipince:
Btw, I definitely think that abc sort is better than non-deterministic, and FWIW, that's also what we do in the dashboard. I send a PR for a server-side implementation here: https://github.com/jetpack-io/axiom/pull/3813

We can drop this one instead. (Note that on the dashboard we also sort client-side, so perhaps now we can drop that).

mikeland73:
The reason I mentioned timestamp is because that seems to be what Vercel does. But agree abc is good, and anything is better than non-deterministic. Server change approved.

PR #267, opened 2024-03-07
[typeid][RFC] Lazy ID parsing

mikeland73:
## Summary

This PR was inspired by our use of stings (instead of typeids) when passing around grpc API data. Because typeids in API requests are encoded as strings, and parsing them every time is tiresome, we end up just passing strings everywhere which defeats the purpose of typeid.

This implements a simple `lazy` type and `Lazy` interface that allows to defer the parsing of the ID until it is actually needed. The main benefit is that it eventually forces error handling, vs when just using strings, we can get away with accepting strings and then using them as strings and never enforcing that they are actually valid typeids.

The main downside to this pattern is that it defers the error handling to later in execution (when parsing is actually needed). This means that any parsing errors may not be caught as soon as someone doing diligent parsing would do. But since we're using strings anyway, we're not actually validating at all. (See example [here](https://github.com/jetpack-io/axiom/blob/2be768c43d33281ef9dd0af01fe0ec1fc064b342/api/midware/auth/authorization.go#L106-L106) (private repo))

## How was it tested?

Unit test


mikeland73:
@loreto thoughts on this?

PR #278, opened 2024-03-29
[updates] Update go versions, api proto models

mikeland73:
## Summary

Mostly automated updates:

* Updated all go versions
* Updated api models (including new tokenservice model)
* Added a few missing dependencies

## How was it tested?

builds, lints


loreto (approved):
LGTM

PR #283, opened 2024-04-01
[filecache] Add IsCacheMiss

mikeland73:
## Summary

In response to:
https://github.com/jetpack-io/devbox/blob/main/internal/nix/search.go#L175-L175


savil (approved):
thanks!

PR #286, opened 2024-04-02
[typeid] Add typeid-unboxed Typescript implementation

LucilleH:
## Summary
This addresses https://github.com/jetpack-io/typeid-js/issues/9

Commit applied: https://github.com/ozanmakes/typeid-unboxed/commit/1d29ba2e59945d1a1a0761b167a3abd1cccbc372

The `TypeID` class now uses the unboxed version as underlying implementation. This is not a breaking change, but rather an addition of unboxed function exposed on top of the class representation.

## How was it tested?
pnpm test


loreto (approved):
LGTM

Although I wish we could just do `typeid()` for the unboxed version (instead of `typeidUnboxed()`, and `TypeID.New()` for the boxed version.

I assume that would be backwards incompatible though?

LucilleH:
> LGTM
> 
> Although I wish we could just do `typeid()` for the unboxed version (instead of `typeidUnboxed()`), and `TypeID.New()` for the boxed version.
> 
> I assume that would be backwards incompatible though?

I mean, we can also do that, and bump the version to 1.0.0

Thoughts?

PR #294, opened 2024-04-09
[easy][api] Update token api

mikeland73:
## Summary

Autogen only


loreto (approved):
LGTM

pkg/api/gen/priv/secrets/v1alpha1/secrets.pb.go
```diff
@@ -20,7 +20,7 @@ const (
 	_ = protoimpl.EnforceVersion(protoimpl.MaxVersion - 20)
 )
 
-// Secret is a resource that represents a Jetify Secret inside a Project.
+// Secret is a resource that represents a Jetpack Secret inside a Project.
```
savil, line 23: undo

savil (approved):

pkg/api/gen/priv/secrets/v1alpha1/secrets.pb.go
loreto: If this is auto-gen it probably means a comment in the .proto file needs to be updated?

PR #317, opened 2024-04-15
[brand] Update jetpack.io -> jetify

mikeland73:
## Summary

Update jetpack.io to jetify.com. Will comment individually on backward compatibility.

### DIFF
--- a/envsec/internal/build/build.go
+++ b/envsec/internal/build/build.go
@@ -33,7 +33,7 @@ func Issuer() string {
 }
 
 func Audience() string {
-	return "https://api.jetpack.io"
+	return "https://api.jetify.com"
 }
 
 func ClientID() string {
--- a/envsec/pkg/awsfed/awsfed.go
+++ b/envsec/pkg/awsfed/awsfed.go
@@ -40,7 +40,7 @@ func (a *AWSFed) AWSCredsWithLocalCache(
 	ctx context.Context,
 	tok *session.Token,
 ) (*types.Credentials, error) {
-	cache := filecache.New[*types.Credentials]("jetify.com/envsec")
+	cache := filecache.New[*types.Credentials]("envsec/aws-creds")
 	return cache.GetOrSetWithTime(
 		cacheKey(tok),
 		func() (*types.Credentials, time.Time, error) {
--- a/envsec/pkg/awsfed/awsfed.go
+++ b/envsec/pkg/awsfed/awsfed.go
@@ -40,7 +40,7 @@ func (a *AWSFed) AWSCredsWithLocalCache(
 	ctx context.Context,
 	tok *session.Token,
 ) (*types.Credentials, error) {
-	cache := filecache.New[*types.Credentials]("jetpack.io/envsec")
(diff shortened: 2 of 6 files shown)


envsec/internal/build/build.go
```diff
@@ -33,7 +33,7 @@ func Issuer() string {
 }
 
 func Audience() string {
-	return "https://api.jetpack.io"
+	return "https://api.jetify.com"
```
mikeland73, line 36: We are currently supporting both audiences so this is backwards compatible.

envsec/pkg/awsfed/awsfed.go
```diff
@@ -40,7 +40,7 @@ func (a *AWSFed) AWSCredsWithLocalCache(
 	ctx context.Context,
 	tok *session.Token,
 ) (*types.Credentials, error) {
-	cache := filecache.New[*types.Credentials]("jetpack.io/envsec")
+	cache := filecache.New[*types.Credentials]("jetify.com/envsec")
```
mikeland73, line 43: This is cache, so no issues changing.

envsec/pkg/envsec/init.go
```diff
@@ -20,7 +20,7 @@ var (
 )
 
 const (
-	dirName       = ".jetpack.io"
+	dirName       = ".jetify.com"
```
mikeland73, line 23: This is gonna cause projects to be reinitialized, but that's OK.

pkg/runx/impl/httpcacher/defaults.go
```diff
@@ -5,7 +5,7 @@ import (
 	"path/filepath"
 )
 
-const xdgSubdir = "jetpack.io/http"
+const xdgSubdir = "jetify.com/http"
```
mikeland73, line 8: This is for caching so OK

pkg/runx/impl/registry/registry.go
```diff
@@ -12,7 +12,7 @@ import (
 	"go.jetpack.io/pkg/runx/impl/types"
 )
 
-var xdgInstallationSubdir = "jetpack.io/pkgs"
+var xdgInstallationSubdir = "jetify.com/pkgs"
```
mikeland73, line 15: This is for caching.

envsec/pkg/envsec/init.go
LucilleH: Should we do `.jetify`?

pkg/runx/impl/httpcacher/defaults.go
LucilleH: Should we do `jetify/http`?

pkg/runx/impl/registry/registry.go
LucilleH: Should we do `jetify/pkgs`?

LucilleH (approved):
Pending QQs

envsec/internal/build/build.go
mikeland73: ory already has both

PR #323, opened 2024-04-20
Update mmzk-typeid (Haskell) compliance to version v0.3

MMZK1526:
https://github.com/MMZK1526/mmzk-typeid/actions/runs/8739099730

## Summary

Update compliance to version v0.3 by extending the typeid prefix syntax according to the new specification.

## How was it tested?

It was tested against the latest provided test case json files.


loreto (approved):
LGTM

PR #324, opened 2024-04-22
update rust type-safe-id version

conradludgate:
## Summary

Updates Rust `type-safe-id` version specification

## How was it tested?

https://github.com/conradludgate/type-safe-id/actions/runs/8665674955/job/23764857568


loreto (approved):
LGTM

PR #332, opened 2024-05-17
Update strong_id's target spec version

johnnynotsolucky:
## Summary

- This PR updates the spec version that `strong_id` 0.4 targets.
- It also corrects the year in @conradludgate's `type-safe-id` package.

## How was it tested?

The Typeid Spec workflow has been updated and is now passing: https://github.com/johnnynotsolucky/strong_id/actions/runs/9133864191


loreto (approved):
LGTM

PR #344, opened 2024-06-26
[typeid-js] Enforce prefix (optionally)

mikeland73:
## Summary

Brings it inline with unboxed version. Makes type-checking much more reliable (because it now throws an error if prefix is wrong). Because the current implementation uses `as T`, if the prefix is not specified, the typechecker type has no relationship to the input data.

Arguably this should have been the default. It requires the same amount of typing because the template type doesn't need to be added:

```ts
const id = TypeID.fromString(stringId, "foo")
// id is of type TypeID<"foo">
```

Unfortunately we can't do this without breaking backward compatibility. 

## How was it tested?

unit tests

### DIFF
--- a/typeid/typeid-js/README.md
+++ b/typeid/typeid-js/README.md
@@ -75,11 +75,9 @@ For example, to parse an existing typeid from a string:
 ```typescript
 import { TypeID } from 'typeid-js';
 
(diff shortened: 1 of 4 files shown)


loreto (approved):
LGTM

PR #346, opened 2024-07-03
[typeid-js] Throw error when parsing typeid with empty suffix

LucilleH:
## Summary
Throw an error when parsing typeid with empty suffix

Addresses https://github.com/jetify-com/typeid-js/issues/12

## How was it tested?
pnpm test

### DIFF
--- a/typeid/typeid-js/src/unboxed/typeid.ts
+++ b/typeid/typeid-js/src/unboxed/typeid.ts
@@ -78,6 +78,10 @@ export function fromString<T extends string>(
     }
   }
 
+  if (!s) {
+    throw new Error(`Invalid TypeId. Suffix cannot be empty`);
+  }
+
   if (prefix && p !== prefix) {
     throw new Error(
       `Invalid TypeId. Prefix mismatch. Expected ${prefix}, got ${p}`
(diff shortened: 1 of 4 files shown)


typeid/typeid-js/src/unboxed/typeid.ts
```diff
@@ -78,6 +78,10 @@ export function fromString<T extends string>(
     }
   }
 
+  if (!s) {
```
mikeland73, line 81: Is this function meant to parse ids missing `_`  anyway?

It seems like line 67 should be:
```ts
if (underscoreIndex === -1)
  throw new Error()
}
```

I see the docbloc, but why does this function allow suffix only? That's not a valid typeid.

mikeland73 (approved):

typeid/typeid-js/src/unboxed/typeid.ts
LucilleH: I think prefix is not required for a typeId. @loreto is that the case?

mikeland73: You are correct. The spec says prefix may be empty and _ omitted: https://github.com/jetify-com/typeid/tree/main/spec

my bad.

PR #349, opened 2024-07-11
[api] Add public apis to new api dir

mikeland73:
## Summary

Same as https://github.com/jetify-com/opensource/pull/348 but only public apis

Adds public api clients for golang and typescript. Adds html and markdown docs.

## How was it tested?

`devbox -c api run gen`


LucilleH (approved):

LucilleH:
Wait this is in opensource repo. Do we want to make all apis public??

PR #351, opened 2024-07-17
[api] Create an unpublished ts api module with example

mikeland73:
## Summary

This creates a new unpublished `@jetify/api` npm package with a simple example.

@LucilleH I was hoping to create this from scratch using best practices (using modules, new eslint flat config format, etc) but could not get it to work so I pretty  much copied our config from `axiom/frontend/packages/api`

Please take a look at:

* package.json
* tsconfig
* I did use new eslint flat config format.

## How was it tested?

`devbox -c api/ts run build`
`devbox -c api/ts run node examples/create-sandbox/index.js`


LucilleH (approved):
Should we name the package `@jetify/client`? cc @loreto

loreto:
> Should we name the package `@jetify/client`? cc @loreto 

I do like that name better. Let's do it.

PR #360, opened 2024-08-01
[ts-api-clint] Update to latest api

mikeland73:
## Summary

All auto-generated

## How was it tested?

Not tested


api/ts/src/gen/pub/sandbox/v1alpha1/sandbox_pb.ts
```diff
@@ -320,6 +336,14 @@ export class ListSandboxesRequest extends Message<ListSandboxesRequest> {
    */
   fetchStatusAndUrl = false;
 
+  /**
+   * Optional. If passed in, only sandboxes for this project are returned.
+   * Project must belong to organization
+   *
+   * @generated from field: string project_id = 2;
+   */
+  projectId = "";
```
LucilleH, line 345: What about createSandbox, can an optional projectId be passed in?

LucilleH (approved):

api/ts/src/gen/pub/sandbox/v1alpha1/sandbox_pb.ts
mikeland73: yep, both `name` and `projectId` are added to create action.

PR #363, opened 2024-08-07
update regex and parse to support multiple underscores in prefix

sjakos:
## Summary

This is an update of the parse and regex matches to support multiple underscores in the prefix for v0.3 of the spec.

### DIFF
--- a/typeid/typeid-sql/sql/03_typeid.sql
+++ b/typeid/typeid-sql/sql/03_typeid.sql
@@ -13,8 +13,8 @@ create or replace function typeid_generate(prefix text)
 returns typeid
 as $$
 begin
-  if (prefix is null) or not (prefix ~ '^[a-z]{0,63}$') then
-    raise exception 'typeid prefix must match the regular expression [a-z]{0,63}';
+  if (prefix is null) or not (prefix ~ '^[a-z_]{0,63}$') then
+    raise exception 'typeid prefix must match the regular expression [a-z_]{0,63}';
   end if;
   return (prefix, uuid_generate_v7())::typeid;
 end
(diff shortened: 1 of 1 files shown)


loreto:
@sjakos Thank you for putting together this PR (and apologies for the delay). Sending comments soon.

typeid/typeid-sql/sql/03_typeid.sql
```diff
@@ -13,8 +13,8 @@ create or replace function typeid_generate(prefix text)
 returns typeid
 as $$
 begin
-  if (prefix is null) or not (prefix ~ '^[a-z]{0,63}$') then
-    raise exception 'typeid prefix must match the regular expression [a-z]{0,63}';
+  if (prefix is null) or not (prefix ~ '^[a-z_]{0,63}$') then
```
loreto, line 16: Technically this is not the right regex expression for two reasons:
1. It allows underscores at the beginning
2. It allows underscores at the end.

But underscores should only be allowed in the middle of the prefix. See https://github.com/jetify-com/typeid/tree/main/spec#type-prefix

The regex suggested there is `^([a-z]([a-z_]{0,61}[a-z])?)?$`

loreto:
In addition to changing the code, we should change the corresponding tests found here: https://github.com/jetify-com/opensource/blob/main/typeid/typeid-sql/supabase/tests/03_typeid.test.sql

To match the latest test cases from the spec:
https://github.com/jetify-com/opensource/blob/main/typeid/typeid/spec/valid.yml
https://github.com/jetify-com/opensource/blob/main/typeid/typeid/spec/invalid.yml

In particular the following test needs to be added to the valid ones:
```
# Tests below were added in v0.3.0 when we started allowing '_' within the
# type prefix.
- name: prefix-underscore
  typeid: "pre_fix_00000000000000000000000000"
  prefix: "pre_fix"
  uuid: "00000000-0000-0000-0000-000000000000"
 ```
 
 This test needs to be removed from the invalid ones:
 ```
 # Test removed in v0.3.0 – we now allow underscores in the prefix
# - name: prefix-underscore
#   typeid: "pre_fix_00000000000000000000000000"
#   description: "The prefix can't have symbols, it needs to be alphabetic"
 ```
 
 And these two tests need to be added to the invalid ones:
 ```
 # Tests below were added in v0.3.0 when we started allowing '_' within the
# type prefix.
- name: prefix-underscore-start
  typeid: "_prefix_00000000000000000000000000"
  description: "The prefix can't start with an underscore"

- name: prefix-underscore-end
  typeid: "prefix__00000000000000000000000000"
  description: "The prefix can't end with an underscore"
 ```

Let me know if you need some guidance adding and running the tests. Alternatively we can check in your code changes (once the regex is updated) and I can send a separate PR for the tests.

PR #364, opened 2024-08-08
[api] update sandboxes list to filter by project-id & member-id

savil:
## Summary

Just ran `cd api && devbox run gen`. Anything else to be done?

Note, this is a breaking change, but we do not expect any current users of this API to be relying on this field yet.
Specifically, we change `project_id` optional field to be be moved into a new `filter` field with `project_id` and `member_id` optional fields.

## How was it tested?

Did not test.


savil:
if it looks good, and tests have passed, feel free to merge upon code review

mikeland73 (approved):
You need to bump version in package.json

PR #368, opened 2024-08-14
[runx] Add support for fetching private GitHub release assets

mikael-lindstrom:
## Summary

Allows runx to download release assets from private repositories, fixes https://github.com/jetify-com/devbox/issues/2223

- Adds a constructor for `download` to set the GitHub token 
- Switches to use the api url instead of browser url when downloading assets
- Sets `Accept` header to `application/octet-stream` to get the binary instead of json metadata for the asset

## How was it tested?

I have tested the `runx` cli locally with and without `RUNX_GITHUB_API_TOKEN` set. With the environment variable the release asset is downloaded correctly from private repositories. Without the environment variable public repositories works fine and private returns `404`.

I also build `devbox` locally and tested with and without `DEVBOX_GITHUB_API_TOKEN` set and it works the same as running the `runx` cli directly.

### DIFF
--- a/pkg/runx/impl/download/download.go
+++ b/pkg/runx/impl/download/download.go
@@ -9,7 +9,17 @@ import (
 	"go.jetpack.io/pkg/runx/impl/fileutil"
 )
 
-func DownloadOnce(url string, dest string) error {
+type Client struct {
+	githubAPIToken string
+}
+
+func NewClient(accessToken string) *Client {
+	return &Client{
+		githubAPIToken: accessToken,
+	}
+}
+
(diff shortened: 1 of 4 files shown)


savil (approved):
LGTM. I also did a rudimentary test by pointing Devbox repo to this PR's changes for runx, and it seemed to work fine. 

@loreto @mikeland73 @LucilleH @gcurtis I have limited experience with `runx` code so tagging you to take a second look. I'll merge this in a short while.

LucilleH:
Looks good to me on the github part 👍

savil:
@mikael-lindstrom thank you!

PR #370, opened 2024-08-16
GHA to publish images to GCP Artifact Registry

ipince:
## Summary
* Adds an action that can build and publish Docker images to GCP's Artifact Registry.
* Adds a job in the `publish-repos` workflow that, for each template, (a) builds and pushes an image to its respective GCP project, and (b) publishes the code to `jetify-examples`.

## How was it tested?
Built and published `hello-go` test template: https://console.cloud.google.com/artifacts/docker/dev-j3tpk/us-east4/github-public/templates%2Fhello-go?authuser=1&orgonly=true&project=dev-j3tpk


action-publish-image/action.yml
```diff
@@ -0,0 +1,69 @@
+name: Build and Push Image
```
savil, line 1: nit about folder organization: Instead of having a top-level `action-publish-image`, could we have a `github-actions/publish-image/` directory structure?

That way we have a place to put other future actions we write here

savil: since we already have `action-publish-code` on top-level, maybe we can move both in a follow up?

savil (approved):

action-publish-image/action.yml
ipince: yep, sounds good, i'll move them in follow-up!

PR #373, opened 2024-09-02
🐛 typeid-js: fix zero not being a valid time

lukas-runge:
Greetings @jetify-com Team, 👋

when running some of my unit tests using time mocking I discovered that a time of `0` currently spawns a `RangeError` in `uuidv7` even tho it's a perfectly valid time. This special case will most likely not occur in production, nevertheless being able to test with mocked time starting from timestamp zero is very convenient. As all data is usually initialized at the beginning of a every test, this would otherwise require initializing system time to some arbitrary value, which sounds more like a workaround than a solution. As this is a limitation of the `uuidv7` library I proposed a fix to @LiosK that would lift this limitation without impacting performance or handling: https://github.com/LiosK/uuidv7/pull/14#issuecomment-2323618094 Sadly he is not interested in lifting this limitation and won't merge the fix.

This PR aims to fix this issue on your end by migrating from https://github.com/LiosK/uuidv7 to https://github.com/uuidjs/uuid which seems to be more widely adopted and better maintained anyway. I also tested performance and found it to be marginally better.

For a a quick reproduction of the issue checkout: https://github.com/lukas-runge/typeid-zero-not-being-a-valid-time 🙌

Best regards,
@lukas-runge


typeid/typeid-js/test/valid.ts
```diff
@@ -17,27 +17,27 @@ export default [
...
-    uuid: "00000000-0000-0000-0000-000000000010",
+    uuid: "00000000-0000-7000-a000-000000000010",
   },
   {
     name: "thirty-two",
-    typeid: "00000000000000000000000010",
+    typeid: "0000000000e00b000000000010",
     prefix: "",
-    uuid: "00000000-0000-0000-0000-000000000020",
+    uuid: "00000000-0000-7000-b000-000000000020",
```
LucilleH, line 40: @lukas-runge can these test cases stay the same, or do you need this to be changed? Prefer that all the test cases stay the same just to be doubly sure this is not a breaking change

LucilleH (commented):
Thanks for the suggestion! Looks reasonable to me.

typeid/typeid-js/test/valid.ts
lukas-runge: @LucilleH, the test cases do need to be changed because they are not compliant with [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-7), the proposed standard defining UUIDv7. For a UUIDv7, the version field must always be set to 0b0111 (hex: 7), and the variant field must be 0b10 (hex: 8, 9, A, or B, as the other two bits are randomly assigned). These changes should not cause any breaking issues unless there is incorrect validation of UUIDv7s somewhere, which would be a bug that needs fixing regardless.

LucilleH (approved):

lukas-runge:
@LucilleH thank you for merging! 🙏 Is there any ETA on when this will be published to npm? 🙌

Best regards,
@lukas-runge

LucilleH:
@lukas-runge published!

PR #377, opened 2024-09-06
[templates] Move active templates to opensource repo

ipince:
## Summary

These are the six templates that are currently "active", i.e., available for use through the dashboard. I just copied the code from their existing repos and removed the `.git` folder.

## How was it tested?
Not tested


LucilleH (approved):
Why do we do this? Is it because we can auto-publish?

ipince:
It's so we can auto-publish a docker image whenever there's a change, and then use that image when deploying a project from a template, for speed.

PR #383, opened 2024-09-09
[cicd] Add action for api/ts and bump versions

mikeland73:
## Summary

* Adds action that checks rules
* bumps node, pnpm, corepack versions


api/ts/package.json
```diff
@@ -48,5 +48,6 @@
   },
   "engines": {
     "node": ">=18"
```
mohsenari, line 50: Do we want to make this ">=20" as well?

mohsenari (approved):

api/ts/package.json
mikeland73: I'm not sure we want to enforce this, as in users should be able to use our library with older node versions. Even this is probably stricter than what we want (and possibly a mistake I introduced when creating this `package.json`)

Since our tsconfig specifies target `ES6` I think we probably support much older versions. Not high priority, but we may want to remove this to allow for wider usage. cc: @LucilleH

PR #384, opened 2024-09-10
[templates] Publish templates from opensource repo

ipince:
## Summary
I'm not 100% sure if this will work the first time, but let's try it. The reason why it might not work is because it'll try to merge whatever is in this repo (in each template) onto the corresponding template repo. It should result in an empty diff and therefore succeed, but again I'm not 100% sure if there's some other check (e.g. history) that might make it fail.

## How was it tested?
After merging I'll fix-forward if needed.


mohsenari (approved):
let's give it a try

PR #396, opened 2024-09-25
Add TypeID community implementation for Erlang

eproxus:
## Summary

Adds keysmith, a new TypeID Erlang community implementation.

## How was it tested?

Markdown has been manually reviewed.


loreto:
@eproxus Thanks for submitting!

Could you point me to the tests validating conformance with the typeid spec? Basically the tests that tests these cases:
- https://github.com/jetify-com/opensource/blob/main/typeid/typeid/spec/invalid.yml
- https://github.com/jetify-com/opensource/blob/main/typeid/typeid/spec/valid.yml

eproxus:
@loreto 👍 I'm working on adding those. I'll let you know when it is done.

eproxus:
@loreto Looking into the test data, it seems the test data does not follow the specification. The specification says:

> 3. A UUID suffix: a 128-bit UUIDv7 encoded as a 26-character string in base32.

But the test data contains (among others) the UUID `00000000-0000-0000-0000-000000000000` which **is not** a valid UUID version 7 (the valid version with all zeroes would be `00000000-0000-7000-8000-000000000000`).

Since my implementation does strict parsing and generation of UUIDs it cannot use this test data for automated tests.

eproxus:
I realize now that any custom UUID is technically allowed by the spec. I've added [validation for the invalid IDs for now](https://github.com/eproxus/keysmith/blob/main/test/keysmith_tests.erl#L169-L179), and will work on adding support for validating the valid IDs as well.

loreto:
> I realize now that any custom UUID is technically allowed by the spec. I've added [validation for the invalid IDs for now](https://github.com/eproxus/keysmith/blob/main/test/keysmith_tests.erl#L169-L179), and will work on adding support for validating the valid IDs as well.

Thanks! Yeah the way I see it is:
- By default a library should always generate new TypeIDs using UUIDv7
- However, if the user of the library so chooses, they can encode any UUID as a TypeID

eproxus:
I've now added support for a more general UUID format so both the valid and invalid test files now pass. You can find the test code here: https://github.com/eproxus/keysmith/blob/main/test/keysmith_tests.erl#L347-L383

loreto:
LGTM

LucilleH (approved):

eproxus:
@loreto Thanks! 🙏

PR #401, opened 2024-10-21
[runx] Fix platform compatibility check for artifacts

edorizzardi:
## Summary
This fixes two errors in the `findArtifactForPlatform` function:
* If the artifact from the GitHub repository is not compatible with the invoker's platform, now it correctly returns a `types.ErrPlatformNotSupported` error
* The way the artifact name string was parsed wasn't working with `x86_64` architecture: the string was split in two because of the underscore, thus never matching correctly. I simplified the logic, by simply checking if the artifact name string, forced to lowercase, contains the invoker's OS and architecture

## How was it tested?
The first issue was tested by creating an ad-hoc GitHub project, compiling it for windows only, and uploading the compiled artifact to a GitHub release, and trying to donwload the binary via both standalone `runx` and with a local `devbox` build from a `linux/amd64` machine.

The second issue was tested by downloading the package `pb33f/wiretap@latest` from GitHub, previously not possible, with both standalone `runx` and a local `devbox` build.

### DIFF
--- a/pkg/runx/impl/registry/artifact.go
+++ b/pkg/runx/impl/registry/artifact.go
@@ -3,16 +3,15 @@ package registry
 import (
 	"path/filepath"
 	"strings"
-	"unicode"
 
 	"go.jetpack.io/pkg/runx/impl/types"
 )
 
 func findArtifactForPlatform(artifacts []types.ArtifactMetadata, platform types.Platform) *types.ArtifactMetadata {
-	var artifactForPlatform types.ArtifactMetadata
+	var artifactForPlatform *types.ArtifactMetadata
 	for _, artifact := range artifacts {
 		if isArtifactForPlatform(artifact, platform) {
-			artifactForPlatform = artifact
+			artifactForPlatform = &artifact
 			if isKnownArchive(artifact.Name) {
 				// We only consider known archives because sometimes releases contain multiple files
 				// for the same platform. Some times those files are alternative installation methods
@@ -23,7 +22,7 @@ func findArtifactForPlatform(artifacts []types.ArtifactMetadata, platform types.
 		}
 	}
 	// Best attempt:
-	return &artifactForPlatform
+	return artifactForPlatform
 }
 
 func isArtifactForPlatform(artifact types.ArtifactMetadata, platform types.Platform) bool {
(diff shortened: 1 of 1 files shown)


mikeland73 (changes requested):
Looks good aside from loop pointer issue. Could you add a few unit tests as well?

Thanks!

pkg/runx/impl/registry/artifact.go
```diff
@@ -3,16 +3,15 @@ package registry
...
 	"go.jetpack.io/pkg/runx/impl/types"
 )
 
 func findArtifactForPlatform(artifacts []types.ArtifactMetadata, platform types.Platform) *types.ArtifactMetadata {
-	var artifactForPlatform types.ArtifactMetadata
+	var artifactForPlatform *types.ArtifactMetadata
 	for _, artifact := range artifacts {
 		if isArtifactForPlatform(artifact, platform) {
-			artifactForPlatform = artifact
+			artifactForPlatform = &artifact
```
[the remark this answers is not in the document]
edorizzardi: @mikeland73 thanks for your review. I added some unit tests, and I reworked the `findArtifactForPlatform` function, so now it returns an error also when the artifact wasn't packaged in a known archive format. Tested it with a release with only a `.deb` artifact.

pkg/runx/impl/registry/artifact.go
```diff
@@ -7,25 +7,33 @@ import (
...
+			// We only consider known archives because sometimes releases contain multiple files
+			// for the same platform. Some times those files are alternative installation methods
+			// like `.dmg`, `.msi`, or `.deb`, and sometimes they are metadata files like `.sha256`
+			// or a `.sig` file. We don't want to install those.
+			knownArchive = true
 		}
+
+		if platformCompatible && knownArchive {
+			return artifact, nil
+		}
```
mikeland73, line 28: Because `platformCompatible` and `knownArchive` get set to true and never get reset it's possible for two different artifacts to make this statement true when combined even if each individual one doesn't actually satisfy. (in practice this is unlikely, but not impossible). You can test this by adding 2 different artifacts, one only satisfies `isArtifactForPlatform` and the other only satisfies `isKnownArchive`.

Here's a small variation on your logic that maintains your goal of splitting no-platform and no-archive errors:

```golang
func findArtifactForPlatform(artifacts []types.ArtifactMetadata, platform types.Platform) (types.ArtifactMetadata, error) {
	foundPlatform := false
	for _, artifact := range artifacts {
		if isArtifactForPlatform(artifact.Name, platform) {
			foundPlatform = true
			if isKnownArchive(artifact.Name) {
				// We only consider known archives because sometimes releases contain multiple files
				// for the same platform. Some times those files are alternative installation methods
				// like `.dmg`, `.msi`, or `.deb`, and sometimes they are metadata files like `.sha256`
				// or a `.sig` file. We don't want to install those.
				return artifact, nil
			}
		}
	}
	if !foundPlatform {
		return types.ArtifactMetadata{}, types.ErrPlatformNotSupported
	}
	return types.ArtifactMetadata{}, types.ErrNoKnownArchive
}
```

mikeland73 (changes requested):
Thanks for the tests! This is very close, one last thing to fix.

pkg/runx/impl/registry/artifact.go
edorizzardi: @mikeland73 oh i see your point. you're right. thank you for the code!

mikeland73 (approved):
🎉 🎉 🎉

PR #441, opened 2025-02-25
Implement `serror` a library for structured errors

loreto:
## Summary
Implement `serror` a library for structured errors

## How was it tested?
devbox run test

All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license](https://www.apache.org/licenses/LICENSE-2.0).
By creating this pull request I represent that I have the right to license the
contributions to the project maintainers under the Apache 2 license.


gcurtis:
How does getting attributes work when handling returned `error`s (not `serror.Error`s)? Won't you have to convert it back to an `serror` with `errors.As`?

```go
func main() {
    err := fail()
    var serr *serrors.Error
    if errors.As(err, &serr) {
        serr.Get("key")
    }
}
```

vs.

```go
func main() {
    err := fail()
    var myerr *MyErrorType
    if errors.As(err, & myerr) {
        myerr.Key
    }
}
```

loreto:
> How does getting attributes work when handling returned `error`s (not `serror.Error`s)? Won't you have to convert it back to an `serror` with `errors.As`?
> 
> ```go
> func main() {
>     err := fail()
>     var serr *serrors.Error
>     if errors.As(err, &serr) {
>         serr.Get("key")
>     }
> }
> ```
> 
> vs.
> 
> ```go
> func main() {
>     err := fail()
>     var myerr *MyErrorType
>     if errors.As(err, & myerr) {
>         myerr.Key
>     }
> }
> ```

As of now, yes. I was thinking that: we'll have a special key that denotes the "type of error". For most use cases, you can then use serror and set the "type" key. I would provide a helper function similar to `errors.As` that can check the "type" and give back the serror. For cases when you don't need to access the values other than for logging, you can use `serror` pretty much like we use `fmt.Errorf` today, but it retains structured data. Ideas on how to improve the API are welcome! (or are you saying having structured errors as defined by this library is not that valuable?)

loreto:
Any further thoughts on this one? @gcurtis @savil

pkg/serror/json.go
```diff
@@ -0,0 +1,194 @@
...
+		if !ok {
+			return nil, false
+		}
+		return int64(n), true
+	case "Uint64":
+		n, ok := groupMap["value"].(float64)
+		if !ok {
+			return nil, false
+		}
+		return uint64(n), true
```
savil, line 170: why does casting to float64 make sense?

savil: according to CoPilot:
> The selection casts the value to float64 because JSON numbers are typically unmarshaled into float64 by default. This ensures that the value can be correctly interpreted and converted to the appropriate type (int64 in this case) during the unmarshaling process.

pkg/serror/serror.go
```diff
@@ -0,0 +1,173 @@
+package serror
```
savil, line 1: reviewed till here. Will come back.

pkg/serror/attr.go
```diff
@@ -0,0 +1,86 @@
...
+//
+// Use Group to collect several key-value pairs under a single
+// key on a log line, or as the result of LogValue
+// in order to log a single value as multiple Attrs.
+func Group(key string, args ...any) Attr {
+	return slog.Group(key, args...)
+}
+
+// Any returns an Attr for the supplied value.
+// See [AnyValue] for how values are treated.
```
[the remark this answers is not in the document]
loreto: AnyValue is defined in slog (which I alias in values.go) – do you think I should be linking to `slog.AnyValue` instead?

pkg/serror/json.go
loreto: bYeah, json unmarshaling will always turn a number into a float64 – but in this case we're adding type that saying that the original marshaled value was an uint64, and therefore we're casting it back. I suppose there's a scenario in which someone purposefully produces a malformed json by hand, and maybe this panics in that case? Do you think I should protect against that?

pkg/serror/serror.go
```diff
@@ -0,0 +1,173 @@
...
+	// TimeKey is the key used for the time when the error was created.
+	// The associated value is a [time.Time].
+	TimeKey = "time"
+)
+
+// Error represents a structured error with attributes that conform to slog conventions.
+// Error is immutable - methods like With() return new instances rather than
+// modifying the original error.
+type Error struct {
+	record slog.Record
```
savil, line 44: clever to use slog.Record here :)

pkg/serror/serror.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// If we have a cause, add it as an error attribute
+	if e.cause != nil {
+		record.AddAttrs(slog.String("cause", e.cause.Error()))
+	}
+
+	// Create a tint handler that writes to our writer
+	handler := tint.NewHandler(w, &tint.Options{
+		Level:   slog.LevelError,
+		NoColor: true,
```
savil, line 116: wait, why use tint if we are not using color?

savil: Did you mean for something like https://github.com/lmittmann/tint?tab=readme-ov-file#automatically-enable-colors ?

savil (approved):
LGTM

Really wish we could _also_ add error-stacks into this

gcurtis:
I guess I'm just wondering what the advantage is of putting the structured data in a `slog.Record`.   Why not just create an error type where its fields are the structured data?

PR #447, opened 2025-03-04
Add additional benchmarks for TypeID

loreto:
## Summary
Adds some additional benchmarks to the existing ones we had in typeid.

## How was it tested?
Ran the benchmarks.

All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license](https://www.apache.org/licenses/LICENSE-2.0).
By creating this pull request I represent that I have the right to license the
contributions to the project maintainers under the Apache 2 license.


typeid/typeid-go/bench_test.go
```diff
@@ -349,6 +371,150 @@ func BenchmarkEncodeDecode(b *testing.B) {
...
+			// Replace a character in the suffix to make it invalid
+			invalidID[len(invalidID)-1] = 'z'
+			invalidIDs[i] = string(invalidID)
+		}
+	}
+
+	b.Run("valid", func(b *testing.B) {
+		b.ReportAllocs()
+		for i := 0; i < b.N; i++ {
+			idx := i % len(validIDs)
```
gcurtis, line 472: I don't think this will be accurate. Each iteration in `b.N` is supposed to do the same thing every time.

typeid/typeid-go/bench_test.go
```diff
@@ -349,6 +371,150 @@ func BenchmarkEncodeDecode(b *testing.B) {
...
+}
+
+// Benchmark mixed operations to simulate real-world usage patterns
+func BenchmarkMixedOperations(b *testing.B) {
+	operations := []string{"create", "parse", "toString", "validate"}
+
+	b.Run("random_mix", func(b *testing.B) {
+		b.ReportAllocs()
+		for i := 0; i < b.N; i++ {
+			op := operations[i%len(operations)]
```
gcurtis, line 499: Same here - each iteration should be the same. A faster machine will have a larger N, so it might run certain operations more than a slower machine.

gcurtis (approved):
Approving, but see my comments about how some of the benchmarks are structured.

PR #480, opened 2025-05-15
Move optional and pointer to opensource, update lint rules

loreto:
## Summary
This PR does two simple changes:
- It makes our linting rules very close to the ones in Axiom
- It moves the code for the optional on-pointer packages from Axiom to open source. This is a straight copy of the code, with no functionality or logic changes

## How was it tested?
Ran unit tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,13 +1,11 @@
-# TODO(Landau): Turn on all linters
-# TODO(Landau): move this to top level of opensource?
 run:
   go: "1.23"
 linters:
   disable-all: true
   enable:
     - dupl
-    - errorlint
     - errcheck
+    - errorlint
     - gofmt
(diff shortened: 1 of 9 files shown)


gcurtis (approved):
Approving because I know this is just for open sourcing something that already exists.

- The pointer package seems to just be `func Ptr[T any](v T) *T { return &v }` which I'm not sure is worth having a new package for.
- The optional package I can see being useful for JSON types, but I'm a little unclear on what some of the semantics are regarding (un)marshalling and nulls vs. field presence.

loreto:
I updated the concurrency comment so that it's not confusing, but I'm going to leave the rest mostly as is since I just want to move the code. I'm open to several of the suggestions, but in follow-up PRs.

PR #489, opened 2025-05-20
add xdg package

savil:
## Summary

This package adds some helper functions for accessing XDG paths for a user

## How was it tested?

`devbox run -- go test -v ./pkg/xdg/...`

### DIFF
--- a/pkg/xdg/xdg.go
+++ b/pkg/xdg/xdg.go
@@ -38,7 +38,7 @@ func resolveDir(envvar, defaultPath string) string {
 
 	home, err := os.UserHomeDir()
 	if err != nil {
-		home = "~"
+		home = "/tmp"
 	}
 
 	return filepath.Join(home, defaultPath)
--- a/pkg/xdg/xdg.go
+++ b/pkg/xdg/xdg.go
@@ -1,0 +1,45 @@
+// Copyright 2025 Jetify Inc. and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+// Package xdg provides access to XDG Base Directory Specification paths.
+package xdg
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func DataSubpath(subpath string) string {
+	return filepath.Join(dataDir(), subpath)
+}
+
+func ConfigSubpath(subpath string) string {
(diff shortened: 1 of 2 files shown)


pkg/xdg/xdg.go
```diff
@@ -0,0 +1,45 @@
...
+
+func resolveDir(envvar, defaultPath string) string {
+	dir := os.Getenv(envvar)
+	if dir != "" {
+		return dir
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		home = "~"
```
gcurtis, line 41: This has been a bug for a while. We should return an error or do something else because non-shells won't expand ~.

gcurtis (approved):
Approving, but not sure when we want to handle fixing `resolveDir`.

loreto:
I thought we already had an xdg package? (although maybe it didn't live in opensource?)

loreto:
Re: returning an error. Are we saying that if we can't resolve home, we want to stop execution of devbox/testpilot? Or are we saying that we should pick a different default directory that is not in home? (in tmp?)

I don't know what the right behavior is, but I wanna double-check that we're not accidentally killing the process with an error when we could continue with a sensible default (if it exists)

savil:
> I thought we already had an xdg package? (although maybe it didn't live in opensource?)

yep, I didn't find it in opensource. This is a port of Devbox's package

savil:
> Re: returning an error. Are we saying that if we can't resolve home, we want to stop execution of devbox/testpilot? Or are we saying that we should pick a different default directory that is not in home? (in tmp?)
> 
> I don't know what the right behavior is, but I wanna double-check that we're not accidentally killing the process with an error when we could continue with a sensible default (if it exists)

Fair point. By returning the errors, most applications will likely fail on this. 

I'm not sure what a good default would be though. Instead of `~`, it could be one of:
- current working dir, but [Getwd](https://pkg.go.dev/os#Getwd) can also return error so back to square one
- `/tmp`
- `/`

Can go with `/tmp` for now

savil:
Ideally the package should return the error and then fallback handling can be added by applications 

But gonna land this for now, for expediency since our apps will require non trivial refactoring to handle at this point

PR #502, opened 2025-05-30
Proposal for v2 of typeid

loreto:
## Summary
Implements a proposed version 2 of TypeID:
+ Removes all the generic stuff and simplifies the API considerably. 
+ Improves the base32 encoder to allow for zero allocations and buffer reuse. 
+ Improved error handling via CorruptInputError and ErrValidation.
+ Adds extensive benchmarks. Rewrote existing ones as well. All benchmarks now use the newer loop notation and use sinks (to prevent compiler optimizations from invalidating the benchmarks)
+ Rethinks the SQL integration to have a nullable version and a non-nullable version. 
+ Changes the internal encoding of the TypeID struct to be more efficient by using a single string and an integer. Note that I tested a byte slice version as well, and it didn't perform as well as this one in the benchmarks. 
+ Improved unit tests so that every single constructor is tested against the spec.
+ Rewrote constructors to minimize allocations as much as possible (some constructors are zero allocations, others are 1 or 2 at most)
+ Changed meaning of IsZero() (so that it makes sense to use with `omitzero`). Introduced HasSuffix() that addresses the same use case as the previous version of IsZero().

Rollout proposal:
+ Merge this PR
+ Update axiom to use it, see if we like it better. Maybe we tweak it a bit more.
+ Once happy with it, we release a new version under `v2`

## How was it tested?
Ran all the tests


typeid/typeid-go/encoding.go
```diff
@@ -25,7 +25,7 @@ func (tid *TypeID[P]) UnmarshalText(text []byte) error {
 
 // MarshalText implements the encoding.TextMarshaler interface.
 // It encodes a TypeID as a string using the same logic as String()
-func (tid TypeID[P]) MarshalText() (text []byte, err error) {
+func (tid TypeID) MarshalText() (text []byte, err error) {
```
gcurtis, line 28: Implement `encoding.TextAppender` and make this `return tid.AppendText(nil)`?

typeid/typeid-go/typeid.go
```diff
@@ -6,42 +6,58 @@ import (
 )
 
 // TypeID is a unique identifier with a given type as defined by the TypeID spec
-type TypeID[P PrefixType] struct {
-	prefix string
-	suffix string
+type TypeID struct {
+	value     string // The complete "prefix_suffix" string
```
gcurtis, line 10: Is the suffix the encoded bytes? Maybe just add a comment to clarify.

typeid/typeid-go/typeid.go
```diff
@@ -6,42 +6,58 @@ import (
...
 }
 
 // String returns the TypeID in it's canonical string representation of the form:
 // <prefix>_<suffix> where <suffix> is the canonical base32 representation of the UUID
-func (tid TypeID[P]) String() string {
-	if tid.Prefix() == "" {
-		return tid.Suffix()
+func (tid TypeID) String() string {
+	if tid.value == "" {
+		return ZeroSuffix
```
gcurtis, line 51: Again, maybe we should adopt the semantics where a zero-value TypeID is invalid and distinct from the ZeroSuffix ID. https://go.dev/play/p/mahjT66CfiM

typeid/typeid-go/typeid.go
```diff
@@ -50,29 +66,43 @@ func (tid TypeID[P]) UUIDBytes() []byte {
...
+//
+// Unlike HasSuffix(), IsZero() returns true only when both:
+// + The prefix is empty (no type specified)
+// + The suffix is the zero suffix "00000000000000000000000000"
+//
+// Only this TypeID would return `IsZero() == true`:
+// + "00000000000000000000000000"
+func (tid TypeID) IsZero() bool {
+	return tid.Prefix() == "" && !tid.HasSuffix()
 }
```
gcurtis, line 101: Is this the behavior we want most of the time? For example, I probably want this to just encode as an empty JSON object:

```go
json.Marshal(&struct{
	ID TypeID `json:"id,omitzero"`
}{}) // I would need to do {ID: typeid.ZeroSuffix} to actually omit the field
```

We're kind of constrained by the IsZero interface here, so the naming is a little strange, but I think we want:

```go
func (tid TypeID) IsZero() bool { return tid.value == "" || tid.Suffix() == ZeroSuffix }
func (tid TypeID) IsValid() bool { return tid.value != "" }
```

gcurtis: Or maybe we just shouldn't check if there's a ZeroSuffix at all. I forget what the original reasoning was for having it.

typeid/typeid-go/base32/base32.go
```diff
@@ -6,13 +6,23 @@ package base32
...
+	"slices"
+	"strconv"
 )
 
 const alphabet = "0123456789abcdefghjkmnpqrstvwxyz"
 
-func Encode(src [16]byte) string {
-	dst := make([]byte, 26)
+// CorruptInputError is returned when invalid base32 data is encountered.
+type CorruptInputError int64
```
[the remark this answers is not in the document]
loreto: Changing to int. I was emulating the encoding library from the standard library ... but in our case we know the exact size of src, so we can change to int.

typeid/typeid-go/constructors.go
```diff
@@ -95,86 +92,102 @@ func split(id string) (string, string, error) {
...
-func FromUUIDBytesWithPrefix(prefix string, bytes []byte) (AnyID, error) {
-	uidStr := uuid.FromBytesOrNil(bytes).String()
-	return FromUUIDWithPrefix(prefix, uidStr)
-}
+// FromUUID encodes the given UUID (in hex string form) as a TypeID with the given prefix.
+// If you want to create an id without a prefix, pass an empty string.
+func FromUUID(uidStr string, prefix string) (TypeID, error) {
+	// Validate prefix early
+	if err := validatePrefix(prefix); err != nil {
+		return zeroID, err
```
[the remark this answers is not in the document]
loreto: The zeroID is valid. 

The question is really whether the "empty" id should default to the zeroid, OR, whether it should represent an illegal ID.

In particular, what happens with these:
```
 // Are these an illegal id by default? Or the zero id by default? 
 
myid := typeid.TypeID{}

var myid typeid.TypeID
```

The current implementation does this: the zero ID is encoded as the empty struct (in fact, if you pass the empty prefix and the 0000000... suffix to any of the constructors, it should detect that special case and end up with the same as `typeid.TypeID{}`)

In fact, the definition of the constant that I'm returning here is: `var zeroID = TypeID{}`


Anyway, long of saying:
+ If we choose to represent the illegal ID as the empty ID, I'll go ahead and change these to `TypeID{}`
+ If we choose to keep the zeroID as == `TypeID{}` then I'll leave as is.

(I'll start a separate discussion to decide on the representation)

typeid/typeid-go/encoding.go
loreto: Good idea. Doing.

typeid/typeid-go/typeid.go
loreto: Doing.

typeid/typeid-go/constructors.go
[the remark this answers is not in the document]
gcurtis: Ah, I see. I mixed up `zeroID` and `ZeroSuffix` in my head. I retract my comment!

As long as `(typeid.TypeID{}).Suffix()` and `(typeid.TypeID{}).String()` return `typeid.ZeroSuffix` (which it looks like they do), then I think the current behavior is ok.

typeid/typeid-go/typeid.go
loreto: Yes! I'll do that.

